PR #1, opened 2021-05-28
track decisions

venkatabhishek:
Overall changes allow for multiple decisions to be added to a scan (i.e. tracking history of decisions similar to scan notes) 
- Remove decision field from scan
- Created annotation model to capture a single decision (fields: creator, created, scan, decision)
- Created annotation view set and updated scan serialization for retrieving decisions
- Created endpoint to add a decision to an existing scan

Note: Included migration for model changes
Note 2: I will create PR for corresponding frontend changes once the new repo is setup


venkatabhishek:
Is there any reason the migration check fails when the migration file is there?

zachmullen:
Normally that would indicate that you modified something in your model after making the migration script. The way to resolve that would typically be to delete the migration file and re-run `makemigrations`, and of course don't forget to `tox -e format` after generating migrations.

miqa/core/models/annotation.py
```diff
@@ -0,0 +1,42 @@
+from enum import Enum
+
+from django.contrib.auth.models import User
+from django.db import models
+from django.utils import timezone
+
+
+class Decision(Enum):
+    NONE = '-'
```
zachmullen, line 9: I guess we still need an explicit `NONE`, if only for the case where an annotation was added, but then someone wishes to explicitly unset it.

miqa/core/rest/scan.py
```diff
@@ -38,10 +42,12 @@ class ScanViewSet(ReadOnlyModelViewSet):
     def decision(self, request, **kwargs):
         serializer = DecisionSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        decision = serializer.validated_data['decision']
         scan = self.get_object()
+        print(serializer.validated_data)
```
zachmullen, line 46: Let's remove this `print`

miqa/core/models/annotation.py
```diff
@@ -0,0 +1,42 @@
...
+        return {
+            cls.NONE.name: '',
+            cls.BAD.name: '0',
+            cls.GOOD.name: '1',
+            cls.USABLE_EXTRA.name: '2',
+        }[decision]
+
+
+class Annotation(models.Model):
+
```
zachmullen, line 34: I believe all our other models are using a UUID primary key, so for consistency let's go ahead and do that here too.

zachmullen (changes requested):
This looks great @venkatabhishek . Just a couple minor details and then re-run the migration.

miqa/core/models/annotation.py
```diff
@@ -0,0 +1,44 @@
...
```
zachmullen, line 35: This is looking good. Let's add `Meta.ordering = ['-created']` and add the composite index for our main query case:

```python
Meta.indexes = [
    models.Index(fields=['scan', '-created']),
]
```

zachmullen (approved):

zachmullen (approved):

zachmullen:
@venkatabhishek when you get a chance, could you resolve conflicts / rebase this?

zachmullen (approved):
LGTM.

Note that this migration will clobber the pending migration on your other branch.

zachmullen (approved):

PR #12, opened 2021-06-02
send bulk resource

venkatabhishek:


miqa/core/models/scan.py
```diff
@@ -40,7 +40,7 @@ class Meta:
         ordering = ['scan_id']
 
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
-    experiment = models.ForeignKey('Experiment', on_delete=models.CASCADE)
+    experiment = models.ForeignKey('Experiment', related_name='scans', on_delete=models.CASCADE)
```
zachmullen, line 43: I believe these changes technically require a DB migration even though they don't (to my knowledge) modify the database.

miqa/core/rest/session.py
```diff
@@ -34,7 +74,9 @@ class Meta:
 
 
 class SessionViewSet(ReadOnlyModelViewSet):
-    queryset = Session.objects.all()
+    queryset = Session.objects.prefetch_related(
+        'experiments__scans__images', 'experiments__scans__notes'
+    )
```
zachmullen, line 79: One design change we should make is to only use this queryset for the `retrieve` action rather than the `list` action. In practice, we'll only want to get this detailed result set for a single session at a time, rather than basically dumping all the data in the system.

We can use `get_queryset` method rather than the `queryset` property to dynamically set the queryset based on the request action.

miqa/core/rest/session.py
```diff
@@ -18,10 +18,50 @@
 from miqa.core.schema.data_import import schema
 
 
+class ImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Image
+        fields = ['id', 'name']
+        ref_name="img 3"
```
zachmullen: Let's change these ref names to be e.g. `'scan_image'`.

zachmullen (changes requested):

miqa/core/rest/session.py
venkatabhishek: I think we would need to change the serializer class as well, because otherwise it will still go and gather foreign key relations. Does something like this look good?
```
    def get_queryset(self):
        if self.action == 'retrieve':
            return Session.objects.prefetch_related(
                'experiments__scans__images', 'experiments__scans__notes'
            )
        else:
            return Session.objects.all()

    def get_serializer_class(self):
        if self.action == 'retrieve':
            return SessionRetrieveSerializer
        else:
            return SessionSerializer
```

zachmullen: Yes, that's exactly what I was thinking 👍

zachmullen (approved):

zachmullen (approved):

PR #19, opened 2021-06-04
add email

### LINKED ISSUE #2: Add endpoint to send emails
The frontend has a feature that allows users to send an email about the current scan. Poke around the [frontend](https://github.com/OpenImaging/miqa-web/blob/master/client/src/components/EmailDialog.vue#L136-L173) to figure out exactly what it's sending to [Girder](https://github.com/OpenImaging/miqa-web/blob/master/server/miqa_server/email.py), then replicate that functionality in Django.

venkatabhishek:
Resolves #2

Important notes:
- Currently, emails are just being printed to console, we would need a running SMTP server or provider to send them
- Screenshot attachment logic assumes images are mime type 'image/*' where * is 'jpeg', 'png', or some other image mime type

### DIFF
--- a/miqa/core/rest/email.py
+++ b/miqa/core/rest/email.py
@@ -2,6 +2,7 @@ from email.mime.image import MIMEImage
 import mimetypes
 import re
 
+from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 from rest_framework import status
 from rest_framework.response import Response
@@ -14,7 +15,7 @@ class EmailView(APIView):
         msg = EmailMultiAlternatives(
             request.data['subject'],
             request.data['body'],
-            'example@kitware.com',  # TODO: replace
+            settings.DEFAULT_FROM_EMAIL,
             request.data['to'],
             bcc=request.data['bcc'],
             cc=request.data['cc'],
--- a/miqa/core/rest/email.py
+++ b/miqa/core/rest/email.py
@@ -24,8 +24,8 @@ class EmailView(APIView):
 
             # parse data uri to extract mime type and base64 data
             # assumptions: mime type is image/jpeg or image/png
-            match = re.search(
-                r'^data:(?P<mime>[\w/\-\.]+);?(\w+),(?P<data>.*)$', screenshot['dataURL']
+            match = re.fullmatch(
+                r'data:(?P<mime>[\w/\-\.]+);?(\w+),(?P<data>.*)', screenshot['dataURL']
             )
 
             if match:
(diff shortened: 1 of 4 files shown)


miqa/core/rest/email.py
```diff
@@ -0,0 +1,45 @@
...
+from django.core.mail import EmailMultiAlternatives
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+
+class EmailView(APIView):
+    def post(self, request, format=None):
+
+        print(request.data['screenshots'])
```
zachmullen, line 14: Remove `print`

zachmullen (changes requested):
LGTM other than the print statement

venkatabhishek:
Maybe not super necessary but I also used a more accurate regex for data urls according to this [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)

miqa/core/rest/email.py
```diff
@@ -26,7 +24,9 @@ def post(self, request, format=None):
 
             # parse data uri to extract mime type and base64 data
             # assumptions: mime type is image/jpeg or image/png
-            match = re.search(r'^data\:(?P<mime>.+);base64,(?P<data>.+)$', screenshot['dataURL'])
+            match = re.search(
+                r'^data:(?P<mime>[\w/\-\.]+);?(\w+),(?P<data>.*)$', screenshot['dataURL']
```
zachmullen, line 28: If you use `re.fullmatch` rather than `re.search` you can omit the `^` / `$`.

zachmullen (approved):

miqa/core/rest/email.py
```diff
@@ -0,0 +1,45 @@
...
+from rest_framework.views import APIView
+
+
+class EmailView(APIView):
+    def post(self, request, format=None):
+
+        msg = EmailMultiAlternatives(
+            request.data['subject'],
+            request.data['body'],
+            'example@kitware.com',  # TODO: replace
```
aashish24, line 17: I believe @scottwittenburg had done something to this end

zachmullen, line 17: We can use the django setting `DEFAULT_FROM_EMAIL` for this. https://github.com/girder/django-composed-configuration/blob/master/composed_configuration/_email.py#L42

zachmullen (changes requested):

dchiquito (approved):

PR #21, opened 2021-06-07
improve development 

venkatabhishek:
- Added `development.md` for new developers (currently only documents docker approach)
- Added django admin command `populate`: Takes as arguments path to csv and username of existing user and populates database with core data; quick access version of import functionality

### DIFF
--- a/miqa/core/management/commands/populate.py
+++ b/miqa/core/management/commands/populate.py
@@ -1,0 +1,23 @@
+from django.contrib.auth.models import User
+import djclick as click
+
+from miqa.core.rest.session import import_data
+from miqa.core.tests.factories import SessionFactory, UserFactory
+
+
+@click.command()
+@click.option('--csv', type=click.Path(exists=True))
+@click.option('--username', type=click.STRING, help='username for session creator')
+def command(csv, username):
+
+    if username:
+
+        user = User.objects.get(username=username)
+
+    else:
+
+        user = UserFactory()
+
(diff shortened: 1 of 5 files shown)


miqa/core/management/commands/populate.py
```diff
@@ -0,0 +1,23 @@
...
+@click.option('--username', type=click.STRING, help='username for session creator')
+def command(csv, username):
+
+    if username:
+
+        user = User.objects.get(username=username)
+
+    else:
+
+        user = UserFactory()
```
zachmullen, line 19: Let's make this use `User.objects.first()` if no username is passed. That will mean, in practice, that it belongs to the superuser created by the developer initially.

miqa/core/management/commands/populate.py
```diff
@@ -0,0 +1,23 @@
...
+
+    session = SessionFactory(name='miqa-dev', import_path=csv, export_path='.', creator=user)
```
zachmullen, line 21:
```suggestion
    session = Session.objects.create(name='miqa-dev', import_path=csv, export_path='.', creator=user)
```

miqa/core/rest/session.py
```diff
@@ -18,6 +18,78 @@
 from miqa.core.schema.data_import import schema
 
 
+def import_data(user, session):
```
zachmullen, line 21: Let's move this to `miqa/core/tasks.py`.

zachmullen (changes requested):

zachmullen (approved):

PR #50, opened 2021-06-24
Adding nn_classifier.py

dzenanz:
This is the current work-in-progress variant of the deep neural network pixel-based classifier. It classifies the images into either good or bad category.

### DIFF
--- a/learning/README.md
+++ b/learning/README.md
@@ -1,0 +1,45 @@
+# MIQA Pixel-based classifier
+
+## Get the program
+```shell
+mkdir nn_work_dir
+cd nn_work_dir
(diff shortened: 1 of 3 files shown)


dzenanz:
@zachmullen do we have some style config file which PyCharm can use to automatically format the source? Most of the failures [here](https://github.com/OpenImaging/miqa/runs/2954099114?check_suite_focus=true) come from wrong code style.

zachmullen:
`tox -e format` will auto-format the code, then `tox -e lint` will run the check locally. The format step can't catch all types of problems; you may need to fix some manually.

dzenanz:
If there is no opposition, I will merge this so the instructions in the readme are valid.

PR #59, opened 2021-07-05
Copy client from miqa-web

dchiquito:
This is a simple copy/paste from miqa-web, nothing really to review here.


brianhelba:
@dchiquito FYI, if you're doing this kind of thing in the future, it'd be preferable to subtree-merge the full Git history of the other repo, instead of copy-pasting all the code. That way, it's possible for future developer-archaeologists to find the true history of this code, rather than everything be attributed to you in this one commit.

See https://alexboffey.co.uk/blog/combining-git-repositories-using-subtree-merging/ for one way to do this.

PR #60, opened 2021-07-06
Production docker compose

dchiquito:
`miqa-web` was already copied into the `client` directory in #59.

* Add a new directory `prod` which contains all the docker/docker-compose/documentation stuff
* The docker image builds the django app, builds the web client, then plops the built artifacts into Django's `staticfiles`. Django's whitenoise app will then serve it under `/static/`. The frontend configuration has been updated to use relative URLs since Django and the web client are now hosted on the same domain/port.
  * To get the `index.html` to be served from `/` rather than `/static/index.html`, I registered `staticfiles` as a template directory, then created a `TemplateView` which served `index.html` and registered it with the url `/`. This will technically break in development, but in development we use `npm run serve` anyway so it doesn't really matter.
* I added a new configuration that does not have any of the extra `composed-configuration` cruft (minio, Sentry, etc.)
* I added a `prod/README.md` that explicitly lists all the setup steps.

Things that are not yet finished:
- [ ] HTTPS. I developed this without the `HttpsMixin`. I ran into problems when trying to log in: after successfully authenticating and authorizing the app, Django refused to redirect me to the app with `ERROR Redirect to scheme 'http' is not permitted`. Including `HttpsMixin` doesn't work locally since nothing is there to handle the SSL termination.
I assume Jim will be using nginx or something to provide SSL termination, so it's possible that everything will just work when he deploys it.
- [ ] A WSGI server. The container work by running `./manage.py runserver 0.0.0.0:8000`. This is likely fine for now, but we should see about installing+running a real WSGI server instead.
- [ ] The logout button wasn't working for me when I tested it, but there were no console or network errors. Something to look into.

### DIFF
--- a/miqa/core/management/commands/makeclient.py
+++ b/miqa/core/management/commands/makeclient.py
@@ -20,8 +20,7 @@ def command(username, uri):
     application = Application(
         name='miqa-client',
         client_id='cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj',
-        client_secret='zNkKPQjc17WfVJjIyLlqpCNdE3LYFfOTF3crpSWsMFMBvxFsqxDz0q85Rgm'
-        'BjYBX18LHB9MrygvqybJ2pmrbkAsSXquHNDVxpiEWxslp6bvQW3gWIYYuvKTyPfWvtwNU',
+        client_secret='',
         client_type='public',
         redirect_uris=uri,
         authorization_grant_type='authorization-code',
--- a/miqa/core/management/commands/makeclient.py
+++ b/miqa/core/management/commands/makeclient.py
@@ -1,0 +1,31 @@
+from django.contrib.auth.models import User
+import djclick as click
+from oauth2_provider.models import Application
+
+
+# create django oauth toolkit appliction (client)
+@click.option('--username', type=click.STRING, help='superuser username for application creator')
+@click.option('--uri', type=click.STRING, help='redirect uri for application')
+@click.command()
+def command(username, uri):
+
+    if username:
+
+        user = User.objects.get(username=username)
+
+    else:
+
+        user = User.objects.first()
+
+    application = Application(
+        name='miqa-client',
+        client_id='cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj',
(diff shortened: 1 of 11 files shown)


miqa/settings.py
```diff
@@ -46,5 +49,25 @@ class ProductionConfiguration(MiqaMixin, ProductionBaseConfiguration):
...
+    """For the production deployment using docker-compose."""
+
+    @staticmethod
+    def before_binding(configuration: ComposedConfiguration) -> None:
+        # Register static files as templates so that the index.html built by the client is
+        # available as a template.
+        # This should be STATIC_ROOT, but that is bound as a property which cannot be evaluated
+        # at this point, so we make this assumption about staticfiles instead.
+        configuration.TEMPLATES[0]['DIRS'] += [
+            str(Path(configuration.BASE_DIR) / 'staticfiles'),
```
zachmullen, line 68: `BASE_DIR` is already a `Path`, and this config value supports paths.

```suggestion
            configuration.BASE_DIR / 'staticfiles',
```

prod/README.md
```diff
@@ -0,0 +1,84 @@
...
+docker-compose run django ./manage.py createsuperuser
+```
+
+### Run the server
+```
+docker-compose up
+```
+
+### Set up OAuth Application
+The app should now be running and accessible in your browser.
```
zachmullen, line 41: I'd like to add an idempotent manage.py command to set up the OAuth application with all these knobs hard-coded to the right value.

zachmullen (changes requested):

venkatabhishek (commented):
Everything seems to work fine, except for the logout button. Clicking it seems to do nothing, but not sure if this behavior is on my end

miqa/core/management/commands/makeclient.py
```diff
@@ -0,0 +1,31 @@
...
+        user = User.objects.get(username=username)
+
+    else:
+
+        user = User.objects.first()
+
+    application = Application(
+        name='miqa-client',
+        client_id='cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj',
+        client_secret='zNkKPQjc17WfVJjIyLlqpCNdE3LYFfOTF3crpSWsMFMBvxFsqxDz0q85Rgm'
```
zachmullen, line 23: IIRC, for the apps we use with SPAs, this secret value can be empty.

zachmullen (changes requested):

PR #61, opened 2021-07-07
Export

dchiquito:
I had to add an `initials` field to `ScanNote`, since we cannot uniquely determine the correct `User` from the initials stored in the CSV. `creator` is still stored and used by default, but `initials` is used as a fallback if it is not available.
Similarly, `Annotation.creator` is now optional since we cannot determine which user made a particular decision from the CSV.

Only one note is imported from the CSV and only the most recent note is exported to CSV.

If a scan has no images, it does not get a row in the exported CSV since we cannot determine the path for that scan.

Other than those caveats, the exported spreadsheet is identical to the import spreadsheet, minus formatting+ordering.


miqa/core/conversion/json_to_csv.py
```diff
@@ -23,7 +23,7 @@ def jsonObjectToCsvContent(jsonObject):
         expId = scan['experiment_id']
 
         scanPath = scan['path']
-        pathComps = scanPath.split(os.path.sep)
+        pathComps = scanPath.rstrip(os.path.sep).split(os.path.sep)
```
zachmullen, line 26: Would all this path manipulation logic be easier if we just made `scanPath` a `pathlib.Path`? Then we could just use its `parts` property.

miqa/core/models/scan_note.py
```diff
@@ -15,6 +15,7 @@ class Meta:
     # We want to set created to arbitrary values during import
     created = models.DateTimeField(default=timezone.now)
     modified = ModificationDateTimeField()
-    creator = models.ForeignKey(User, on_delete=models.PROTECT)
+    creator = models.ForeignKey(User, on_delete=models.PROTECT, null=True, blank=True)
+    initials = models.CharField(max_length=2)
```
zachmullen, line 19: It looks like the default value listed in the migration is different from the default value listed here. I don't really get how the checkmigrations step is passing... what gives?

miqa/core/tasks.py
```diff
@@ -90,3 +98,78 @@ def import_data(user, session):
...
+        'sites': [{'name': site} for site in sites],
+    }
+
+    if session.export_path.endswith('.csv'):
+        with open(session.export_path, 'w') as csv_file:
+            csv_content = jsonObjectToCsvContent(json_content)
+            csv_file.write(csv_content.getvalue())
+    elif session.export_path.endswith('.json'):
+        with open(session.export_path, 'w') as json_file:
+            json_file.write(json.dump(json_content))
```
zachmullen, line 175: `else`? 

Should we detect this early and fail, or have we already validated that it's one of these two? (If the latter, let's just change the second condition to `else`.

miqa/core/models/scan_note.py
zachmullen, line 19: At first glance I thought this was just a denormalization that could be replaced with

```python
@property
def initials(self) -> str:
    if self.creator:
        return f'{self.creator.first_name[0]}{self.creator.last_name[0]}'
    return '??'
```

But now I see that the initials are able to be imported, so this model makes sense. Carry on.

zachmullen (changes requested):

curtislisle:
Thanks for these improvements.  In the commit note, it was mentioned that
only the most recent note is exported to CSV.  Are multiple notes stored in
the Django model, or is an existing note replaced when a user saves a new
note?  Thanks.

miqa/core/models/scan_note.py
dchiquito: The migration default is only necessary because it's adding a non-null field with no default. When the migration is applied something needs to be put into the new column for existing rows, so I just decided it would be `??`.

zachmullen: Oh I see, that makes sense, thanks.

miqa/core/tasks.py
venkatabhishek, line 175:
```suggestion
            json_file.write(json.dumps(json_content, indent=4))
```
`json.dump` requires a file pointer, `json.dumps` with indent keywordarg will pretty-print

client/src/components/DataImportExport.vue
```diff
@@ -37,11 +37,11 @@ export default {
...
-      // this.$prompt({
-      //   title: 'Export',
-      //   text: 'Saved data to file successfully.',
-      //   positiveButton: 'Ok',
-      // });
+      await this.djangoRest.export(this.mainSession.id);
+      this.$snackbar({
+        text: 'Saved data to file successfully.',
+        timeout: 6000,
+      });
```
venkatabhishek, line 44: Should we include some sort of feedback/loading icon? There is no visual change after pressing the export button until the snackbar is shown. Also, if an error occurs in the export process, no visual update is made.

venkatabhishek (changes requested):

dchiquito:
@curtislisle Multiple notes are stored in the Django model, I just wasn't sure how to import/export multiple notes in a single cell in a CSV.

I just checked the girder implementation again, and it uses unix CSV encoding, which wraps every cell in double quotes. This allows new lines `\n` in cells, which is how it encoded multiple notes. I will update this PR accordingly.

client/src/components/DataImportExport.vue
dchiquito: I'm only trying to achieve feature parity here, although that would be an excellent UX improvement. I think it would be better as a separate PR.

PR #65, opened 2021-07-14
Secret management support for prod environment

zachmullen:
This change works for setting up a postgres password, and has a generally better pattern for env var management.

~~It's a draft for the moment because I'm running into a weird issue where the django image build isn't receiving the right env vars and I'm not sure why.~~


prod/README.md
```diff
@@ -4,14 +4,17 @@ This directory contains everything required to deploy MIQA in production using d
...
 
-You can also specify those variables explicitly in your local shell environment instead of editing the file.
-`docker-compose` will check for local shell values before defaulting to the values in `.env`.
+`cp .env.template .env`
+
+Open the `.env` file and make changes to the variables in the top section.
+
+* `DJANGO_EMAIL_URL` must be of the form `submission://username:password@my.smtp.server:port`. See the [documentation](https://github.com/migonzalvar/dj-email-url#supported-backends) for details with encoding special characters.
+
+You can also specify those variables explicitly in your local shell environment instead of editing the file. `docker-compose` will check for local shell values before defaulting to the values in `.env`.
```
venkatabhishek, line 17: Just checking, but  `docker-compose` won't override variables in the local shell with values in `.env` correct? Will it only default to values in `.env` if they are undefined/empty?

zachmullen: I actually haven't tested this. I thought about deleting it but wasn't sure. Maybe I'll just delete it.

dchiquito: > Values in the shell take precedence over those specified in the .env file.

\- https://docs.docker.com/compose/environment-variables/

dchiquito (approved):

PR #68, opened 2021-07-15
Allow non-root production deployments

dchiquito:
Set `DJANGO_MIQA_URL_PREFIX` to something like `/miqa2` to serve from `.../miqa2/` instead of the domain root.

I also included an example nginx configuration.


miqa/settings.py
```diff
@@ -58,6 +60,33 @@ class DockerComposeProductionConfiguration(
 ):
     """For the production deployment using docker-compose."""
 
+    MIQA_URL_PREFIX = values.Value(environ=True, default='/')
+
+    @property
+    def STATIC_URL(self):
+        """Prepend the MIQA_URL_PREFIX to STATIC_URL."""
+        return str(Path(self.MIQA_URL_PREFIX) / 'static') + '/'
```
zachmullen, line 68:
```suggestion
        return f'{Path(self.MIQA_URL_PREFIX) / "static"}/'
```

zachmullen (changes requested):

venkatabhishek (approved):

Issue #46, opened 2021-06-18
Require users to log in for all endpoints


dchiquito:
Should be simply adding a permission requirement to all of the ViewSets.

Setting import/export path should be an admin action.

zachmullen:
I realize I'm a bit late to this party, but I think we should set DEFAULT_PERMISSION_CLASSES in our MiqaMixin config, overriding [this inherited default](https://github.com/girder/django-composed-configuration/blob/master/composed_configuration/_rest_framework.py#L75).

PR #69, opened 2021-07-19
fix permissions/login error

venkatabhishek:
Change front-end to match updated back-end permissions: previously front-end was making requests before outh login occured, thus preventing entire site from loading

- Also remove unused `user` field in `djangoClient`
- Fix itk poolsize error
- Update default permissions class as per https://github.com/OpenImaging/miqa/issues/46#issuecomment-881462009

### DIFF
--- a/miqa/settings.py
+++ b/miqa/settings.py
@@ -38,6 +38,9 @@ class MiqaMixin(ConfigMixin):
         # oauth session
         configuration.OAUTH2_PROVIDER['ACCESS_TOKEN_EXPIRE_SECONDS'] = 1800  # 30 minutes
 
+        # drf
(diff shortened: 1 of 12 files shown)


miqa/settings.py
```diff
@@ -38,6 +38,9 @@ def before_binding(configuration: ComposedConfiguration) -> None:
         # oauth session
         configuration.OAUTH2_PROVIDER['ACCESS_TOKEN_EXPIRE_SECONDS'] = 1800  # 30 minutes
 
+        # drf
+        configuration.REST_FRAMEWORK['DEFAULT_PERMISSION_CLASSES'] = ['rest_framework.permissions.IsAuthenticated']
```
dchiquito, line 42: Does this mean that you no longer need to specify `IsAuthenticated` on every view?

venkatabhishek: Yup I will make this change

dchiquito (approved):

PR #76, opened 2021-07-26
Disable SSL redirect and set SECURE_PROXY_SSL_HEADER

dchiquito:
The reverse proxy was doing SSL termination and sending HTTP traffic to
django, which was then doing SSL redirection instead of handling the
traffic.

* SECURE_SSL_REDIRECT is set to False to stop this behavior.
* SECURE_PROXY_SSL_HEADER is set to ('HTTP_X_FORWARDED_PROTO', 'https')
so that when the reverse proxy sets the `X-Forwarded-Proto` header,
Django will treat the HTTP traffic as HTTPS.

The nginx example is also update to set `X-Forwarded-Proto`.

### DIFF
--- a/miqa/settings.py
+++ b/miqa/settings.py
@@ -56,7 +56,6 @@ class ProductionConfiguration(MiqaMixin, ProductionBaseConfiguration):
     pass
 
 
-# TODO include HttpsMixin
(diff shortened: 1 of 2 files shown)


zachmullen (approved):
Excellent, thanks!

PR #77, opened 2021-07-26
Session Lock UI

### LINKED ISSUE #75: Front-end code for session locks
This is  a follow up to #72 . To isolate the frontend and backend work, we'll do them as separate PRs.

venkatabhishek:
Resolve #75 

- Move `mainSession` from inject to vuex store: Allows for changing `mainSession` (specifically `lock_owner`) more easily as well as for eventual multiple sessions
- Add section in nav bar that displays name + lock ownership for current session. Corresponding actions are:
    - Current owner: Release
    - No owner: Acquire
    - Owned by someone else: No action available

### DIFF
--- a/client/src/components/SessionLock.vue
+++ b/client/src/components/SessionLock.vue
@@ -15,13 +15,7 @@ export default {
         await this.djangoRest.acquireSession(this.mainSession.id);
         this.setMainSession({ ...this.mainSession, lock_owner: this.user });
       } catch (err) {
-        if (err.response && err.response.status === 409) {
-          this.updateCurrentSession();
-          this.$snackbar({
-            text: 'The lock is held by a different user.',
-            timeout: 6000,
-          });
-        } else {
+        if (err.response && err.response.status !== 409) {
           throw err;
         }
       }
@@ -31,13 +25,7 @@ export default {
         await this.djangoRest.releaseSession(this.mainSession.id);
         this.setMainSession({ ...this.mainSession, lock_owner: null });
       } catch (err) {
-        if (err.response && err.response.status === 409) {
-          this.updateCurrentSession();
-          this.$snackbar({
-            text: 'The lock is held by a different user.',
-            timeout: 6000,
-          });
-        } else {
+        if (err.response && err.response.status !== 409) {
           throw err;
         }
       }
--- a/client/src/components/SessionLock.vue
+++ b/client/src/components/SessionLock.vue
@@ -1,5 +1,5 @@
 <script>
-import { mapState, mapMutations } from 'vuex';
+import { mapState, mapActions, mapMutations } from 'vuex';
 
 export default {
   name: 'SessionLock',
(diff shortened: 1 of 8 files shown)


client/src/components/SessionLock.vue
```diff
@@ -0,0 +1,74 @@
...
+    ...mapState(['mainSession']),
+  },
+  methods: {
+    ...mapMutations(['setMainSession']),
+    async acquireLock() {
+      try {
+        this.djangoRest.acquireSession(this.mainSession.id);
+        this.setMainSession({ ...this.mainSession, lock_owner: this.user });
+      } catch (err) {
+        console.log(err);
```
zachmullen, line 17: We'll want to handle this more gracefully, as it's a situation that we should reasonably expect to occur -- two users trying to lock the session at roughly the same time.

client/src/components/SessionLock.vue
```diff
@@ -0,0 +1,74 @@
...
+      }
+    },
+    async releaseLock() {
+      try {
+        this.djangoRest.releaseSession(this.mainSession.id);
+        this.setMainSession({ ...this.mainSession, lock_owner: null });
+      } catch (err) {
+        console.log(err);
```
zachmullen, line 25: This has the effect of ignoring this error. It's generally preferable, in the case of behaviorally unhandled exceptions, to let them bubble up out of the stack. That will log them to the console anyway, and if we end up using Sentry or some other similar tool, we'll get all the nice features of telemetry, logging, and alerting.

zachmullen (approved):

zachmullen (changes requested):
See inline comments

client/src/components/SessionLock.vue
venkatabhishek: One idea I had was to just show a snack bar with message "This session is already owned" upon status 409 (i.e. someone else has acquired the lock in the time that the user was active and the endpoint raised `LockContention`). How should we update the local state of the lock owner when this occurs? Could we update `LockContention` to include the current lock owner or should we make another request to `/sessions` in order to update the local state?

zachmullen: I think ultimately we'll want to hit an endpoint that refreshes the lock_owner. The way we've got the API set up now, retrieving the session is really expensive, so I think we'll need to either have a different endpoint for just the lock owner, or else move our existing expensive session retrieval endpoint. I'm leaning toward the latter, maybe move that from `GET /sessions/{id}` to `GET /sessions/{id}/deep` or something, so that a shallow lookup can be performed cheaply whenever we want.

client/src/django.js
```diff
@@ -51,9 +51,19 @@ const djangoClient = new Vue({
       return results;
     },
     async session(sessionId) {
+      const { data } = await apiClient.get(`/sessions/${sessionId}/deep`);
+      return data;
+    },
+    async sessionDetail(sessionId) {
```
dchiquito, line 57: I'd assume that `sessionDetail` would have more information than `session`, but it seems to be the other way round.

zachmullen: Detail is the right term of art in DRF-land, but I agree it's confusing next to the name `session` here. Maybe we can just name the other method `sessionDetailDeep` if we want to keep this name.

venkatabhishek: Changed names to 
`session`: shallow request (i.e. just id, name, and lock_owner)
`sessionDeep` deep request (i.e. all above + hierarchy of experiments, scans, etc.)

PR #79, opened 2021-07-27
Fix screenshots

dchiquito:
I found a way to piggy back on our existing view generators so I didn't have to figure out how to respecify the whole render pipeline. The `ScreenshotView2D_x/y/z` are copies of `View2D_x/y/z`, but are not mounted in the DOM. Instead, when the screenshot button is clicked, they are mounted in a disconnected div, sized appropriately, and used to capture the screenshot.


aashish24:
@dchiquito can you post some images before and after?

dchiquito:
@aashish24 Two screenshots taken and downloaded with the window size as small as I could get it while still being able to click the screenshot button:
Before:
small-window-bugged
After:
small-window-fixed
Email is harder to generate when running locally, but I could figure that out if necessary.

PR #81, opened 2021-07-28
Add yarn lint to GitHub Actions

dchiquito:
We are now running `tox` in CI, but since the client code also lives in CI, we should also be running `yarn lint`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,3 +47,16 @@ jobs:
           DJANGO_MINIO_STORAGE_ENDPOINT: localhost:9000
           DJANGO_MINIO_STORAGE_ACCESS_KEY: minioAccessKey
           DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
+  test-ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: client
+    steps:
+      -
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -47,3 +47,16 @@ jobs:
           DJANGO_MINIO_STORAGE_ENDPOINT: localhost:9000
           DJANGO_MINIO_STORAGE_ACCESS_KEY: minioAccessKey
           DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
+  test-ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: client
+    steps:
+      -
```
zachmullen, line 56: ^ This guy right here

dchiquito:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -47,3 +47,15 @@ jobs:
           DJANGO_MINIO_STORAGE_ENDPOINT: localhost:9000
           DJANGO_MINIO_STORAGE_ACCESS_KEY: minioAccessKey
           DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
+  test-ci:
```
dchiquito, line 50:
```suggestion
  test-client:
```

dchiquito:
The error is fixed in https://github.com/OpenImaging/miqa/pull/79/commits/58c844ac2fe0514f5f38b24ef75180a59ce1eb33, I'm gonna merge this

PR #100, opened 2021-08-16
Task list

venkatabhishek:
- Create **Task** model: A task consists of multiple experiments and is assigned to a single user
- Create new tab to view/switch sessions and tasks
- Add task model to admin page (currently the only way to create/assign tasks)
- Switched session state from inject to vuex store as it now requires modification 

Tasks are loaded into the viewer similar to the local load feature: they are grouped into a single "dummy" session and placed into the vuex store

Added a mode field to vuex store. When mode is 'SESSION', application assumes that an entire session will be loaded into viewer. When mode is 'TASK', application assumes vuex store is already populated with all corresponding model meta data.


miqa/core/rest/task.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class TaskViewSet(GenericViewSet):
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+    
+    permission_classes = [IsAuthenticated]
+
+    @action(detail=False, pagination_class=None)
+    def me(self, request):
```
dchiquito, line 28: If this is the only method on the ViewSet, I don't see any reason not to make this `retrieve`. Is there some future use case where a supervisor/admin would want to see every task in the system (/tasks), but users would want to use a different endpoint (/tasks/me) to see only their own tasks?

miqa/core/rest/task.py
```diff
@@ -0,0 +1,33 @@
...
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+
+    permission_classes = [IsAuthenticated]
+
+    @action(detail=False, pagination_class=None)
+    def me(self, request):
+        """Return the currently logged in user's tasks."""
+        queryset = Task.objects.all()
+        queryset = queryset.filter(user=request.user)
```
dchiquito, line 31:
```suggestion
        queryset = Task.objects.filter(user=request.user)
```

client/src/views/TaskList.vue
```diff
@@ -0,0 +1,139 @@
...
+  },
+  inject: ['djangoRest', 'user'],
+  async created() {
+    this.sessions = await this.djangoRest.sessions();
+    this.tasks = await this.djangoRest.tasks();
+  },
+  methods: {
+    ...mapMutations(['setMainSession', 'setMode', 'setDrawer']),
+    ...mapActions(['loadTask']),
+    mapExperiments(experiments) {
```
dchiquito, line 27:
```suggestion
    experimentNames(experiments) {
```

client/src/store/index.js
```diff
@@ -503,15 +502,13 @@ const store = new Vuex.Store({
         });
 
         // Web sessions == Django scans
```
dchiquito, line 504:
```suggestion
```

client/src/store/index.js
```diff
@@ -503,15 +502,13 @@ const store = new Vuex.Store({
...
-        // TODO these requests *can* be run in parallel, or collapsed into one XHR
         // eslint-disable-next-line no-await-in-loop
         const { scans } = experiment;
         for (let j = 0; j < scans.length; j += 1) {
           const scan = scans[j];
           commit('addExperimentSessions', { eid: experiment.id, sid: scan.id });
 
           // Web datasets == Django images
```
dchiquito, line 511:
```suggestion
```

dchiquito (commented):
I'm not clear at all on what `mode` means. The only place I see it referenced is that all the session experiments are loaded when loading a dataset in `SESSION` mode. Are there changes coming soon that will use mode in the UI somehow?

miqa/core/rest/task.py
venkatabhishek: This is exactly what I was anticipating. This action gets the tasks for the current user and I assumed a supervisor view would want to be able to see/filter all tasks. If this use case does not exist then I can make this change

venkatabhishek:
@dchiquito The mode is used to ensure that the vuex state is not overwritten when loading a task (i.e. multiple specific experiments) into `Dataset.vue`. The flow works as follows: 
1. User navigates to tasks tab and selects a task
2. Selecting a task loads the associated experiments into the vuex state as a dummy session
3. User is redirected back to `Dataset.vue` at uri `/` 

Previously whenever the user navigates to `/`, `Dataset.vue` resets the vuex state and loads the session to populate the vuex state. However, because we are using a task we don't want to overwrite the vuex state (it has already been populated in the `loadTask` vuex action). Thus, the mode helps differentiate when we are loading a session or a task. Does this make sense?

dchiquito:
@venkatabhishek I think it might make more sense to redesign the router a bit so we don't have to use that workaround. Right now it looks like the only relevant router entry is:
```
    {
      path: '/:datasetId?',
      name: 'dataset',
      component: Dataset,
    },
```
I believe `dataset` refers to a scan. This makes sense when we are implicitly using the main session, but breaks down when there are multiple contexts to view a scan from. I think we could reasonably break this into `/session/:sessionId/scan/:datasetId` and `/task/:taskId/scan/:scanId`. We would probably also have to rename the `session` in Vuex to something that reflects that it could contain the session contents or a tasks contents, or we could refactor Vuex to store those things separately.

This sounds simpler to me than having to set and get a mode at different points while the dataset is being loaded. What do you think?

venkatabhishek:
@dchiquito Yes that sounds a lot better. One question I have on this approach is then: how do we separate out loading a session vs a task? 

For example, do we route both of those uris to `Dataset.vue` and pass a prop via the router like:
```
{
  path: '/session/:sessionId/scan/:datasetId',
  name: 'dataset',
  component: Dataset,
  props: {
    mode: 'session'
  }
},
{
  path: '/task/:taskId/scan/:scanId',
  name: 'task',
  component: Dataset,
  props: {
    mode: 'task'
  }
}
```
and then have some conditional logic in `Dataset.vue` that reads `props.mode` to load (i.e. make a request for) a session or task?

Or should we refactor out the shared components in `Dataset.vue` and have a `Session.vue` and `Task.vue` that have their respective loading in `created` hook?

dchiquito:
@venkatabhishek `Dataset.vue` looks ripe for a refactoring into smaller chunks, but that's probably best undertaken in a separate PR (probably after your departure :cry:). 

I think for now something like this would work:
```
// router.js
{
  path: '/session/:sessionId/scan/:scanId',
  name: 'dataset',
  component: Dataset,
},
{
  path: '/task/:taskId/scan/:scanId',
  name: 'task',
  component: Dataset,
}

// Dataset.vue
  async created() {
    const { sessionId, taskId, scanId } = this.$route.params;
    if (sessionId) { ... load session ... }
    else if (taskId) { ... load task ... }
    ...
}
```
The `mode` selector thing smells like an antipattern to me, although I don't have any sources to cite.

aashish24:
We decided to go different route hence closing this issue

PR #101, opened 2021-08-18
Fix logout

dchiquito:
In Girder 4, there are actually two "logins" in play: logging in to Django and the OAuth token the frontend uses to communicate with the Django API. The web GUI discards the OAuth token on logout or when the session expires, but the user still has a session cookie with Django. Django would see the unexpired session cookie and immediately issue a new OAuth token, which defeats the purpose of logging out in the first place.

This PR sets `SESSION_COOKIE_AGE` to 5 seconds, which is very short while hopefully still lasting long enough for Django to complete the OAuth flow, even under server load.

Also, create a custom Admin site that explicitly expires in 30 minutes when logging in to `/admin`. The default behavior is to use `SESSION_COOKIE_AGE` as the admin session length, which is now obviously too short.

### DIFF
--- a/miqa/core/admin.py
+++ b/miqa/core/admin.py
@@ -4,6 +4,7 @@ from django.contrib.admin.forms import AdminAuthenticationForm
 
 from .models import Annotation, Experiment, Image, Scan, ScanNote, Session, Site
 
+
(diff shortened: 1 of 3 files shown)


dchiquito:
Zach pointed out a solution that will work better.

PR #105, opened 2021-08-19
Logout from django

dchiquito:
* Add a view to explicitly log out from Django. There is an existing form at `/accounts/logout`, but it requires CSRF and the client is currently only set up to send requests to `/api/v1/...`, so it is easiest to just create a new view.
* Use that view to explicitly log out from Django when revoking the OAuth token.
* Expire the Django session after 30 minutes.

### DIFF
--- a/client/src/django.js
+++ b/client/src/django.js
@@ -37,6 +37,7 @@ const djangoClient = new Vue({
       await oauthClient.redirectToLogin();
     },
     async logout() {
+      await apiClient.post('/logout/', undefined, { withCredentials: true });
       await oauthClient.logout();
     },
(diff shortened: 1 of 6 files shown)


client/src/django.js
```diff
@@ -37,6 +37,7 @@ const djangoClient = new Vue({
       await oauthClient.redirectToLogin();
     },
     async logout() {
+      await apiClient.post('/logout/', undefined, { withCredentials: true });
```
zachmullen, line 40: I'm not sure how this will hold up long term, but this seems good enough for now. 👍

zachmullen (approved):

PR #106, opened 2021-08-25
Disable session locks

dchiquito:
Right now in master, users are unable to perform imports or do any other write operation without the session lock, which currently does not have a UI or any descriptive error messages on how to obtain the session lock.

Remove `UserHoldsSessionLock` from the permissions of all the views for now so that `master` is usable.

### DIFF
--- a/miqa/core/rest/annotation.py
+++ b/miqa/core/rest/annotation.py
@@ -5,8 +5,6 @@ from rest_framework.viewsets import GenericViewSet
 
 from miqa.core.models import Annotation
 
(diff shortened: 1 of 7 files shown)


zachmullen (approved):

zachmullen:
Are we intending to bring this functionality back at some point? I'm cool with this changeset if that's the case.

dchiquito:
I'm planning on moving Abhi's work into a feature branch that can be worked on without breaking master.

PR #112, opened 2021-08-31
Add an initial setup command to run npm install

annehaley:
We discovered that this command needs to be run before the first installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,14 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker-compose run --rm django ./manage.py migrate`
-2. Run `docker-compose run --rm django ./manage.py createsuperuser`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,13 +4,14 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker-compose run --rm django ./manage.py migrate`
-2. Run `docker-compose run --rm django ./manage.py createsuperuser`
+1. Run `docker-compose run --rm npm npm install`
```
dzenanz, line 7: Should there be `npm npm` (npm repeated twice)?

annehaley: yes, @zachmullen explained that the first `npm` is the service and `npm install` is the command to run:
https://docs.docker.com/compose/reference/run/

zachmullen (approved):

PR #114, opened 2021-08-31
Typescript

dchiquito:
I recommend browsing by commit as this is a rather chunky PR.

* Install and configure TypeScript
* Convert `constants.js`, `django.js`, `main.js` and `store/index.js` to TypeScript
* Install direct-vuex and convert `store/index.ts` to use it
* Install composition-api
* Fix the major TypeScript bugs that block the production build

Currently, `strict: false` in `tsconfig.json` because there are many, many TypeScript warnings to squash. If `strict: true`, then the production build fails. We should aim to eventually turn strict on, but to merge this I think it needs to be off.


zachmullen:
Is the CI failure expected?

client/src/store/index.ts
```diff
@@ -1,4 +1,5 @@
-import Promise from 'bluebird';
+import BluebirdPromise from 'bluebird';
+import { createDirectStore } from "direct-vuex"
```
zachmullen, line 2: Does our linter enforce quote characters and semicolons on these?

client/src/store/index.ts
```diff
@@ -66,7 +67,7 @@ function getArrayName(filename) {
 }
 
 function getData(id, file, webWorker = null) {
-  return new Promise((resolve, reject) => {
+  return new BluebirdPromise((resolve, reject) => {
```
zachmullen, line 70: Do you know why we were even using bluebird in the first place?

dchiquito: No idea, we could maybe ask Scott. Rather than figure it out/exorcise it I just did the minimum to stop the TS errors.

client/src/store/index.ts
dchiquito: It does, I just forgot to run it :/

client/src/store/index.ts
zachmullen: Ok, I'll file a ticket. It doesn't look like it's required for either itk.js or vtk.js, so hopefully we can rip it out.

annehaley (approved):
We went over most of these changes commit-by-commit on a call. This looks like a good start to the ts conversion; I should be able to pick this up from here by doing more of the same kinds of changes. I'll be referencing these commits as I move forward.

PR #132, opened 2021-09-02
Redirect / to localhost:8081 in dev mode

### LINKED ISSUE #113: Redirect / to localhost:8081 in dev mode

annehaley:
Resolves #113.

### DIFF
--- a/miqa/urls.py
+++ b/miqa/urls.py
@@ -1,6 +1,9 @@
+import os
+
 from django.conf import settings
 from django.contrib import admin
 from django.urls import include, path
+from django.views.generic.base import RedirectView
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions, routers
@@ -47,6 +50,11 @@ urlpatterns = [
     path('api/docs/redoc/', schema_view.with_ui('redoc'), name='docs-redoc'),
     path('api/docs/swagger/', schema_view.with_ui('swagger'), name='docs-swagger'),
 ]
+
+if os.environ['DJANGO_CONFIGURATION'] == 'DevelopmentConfiguration':
+    urlpatterns = [
+        path('', RedirectView.as_view(url='http://localhost:8081'), name='home-redirect')
+    ] + urlpatterns
 
 if settings.DEBUG:
     import debug_toolbar
--- a/miqa/urls.py
+++ b/miqa/urls.py
@@ -1,9 +1,6 @@
-import os
-
 from django.conf import settings
 from django.contrib import admin
 from django.urls import include, path
-from django.views.generic.base import RedirectView
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions, routers
@@ -40,7 +37,7 @@ schema_view = get_schema_view(
 )
 
 urlpatterns = [
-    path('', HomePageView.as_view(), name='home'),
+    path('', HomePageView, name='home'),
     path('accounts/', include('allauth.urls')),
     path('oauth/', include('oauth2_provider.urls', namespace='oauth2_provider')),
     path('admin/', admin.site.urls),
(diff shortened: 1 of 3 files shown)


miqa/urls.py
```diff
@@ -48,6 +51,11 @@
     path('api/docs/swagger/', schema_view.with_ui('swagger'), name='docs-swagger'),
 ]
 
+if os.environ['DJANGO_CONFIGURATION'] == 'DevelopmentConfiguration':
```
zachmullen, line 54: Rather than check the configuration, let's make an explicit setting called `HOMEPAGE_REDIRECT_URL`. We should add it to the config mixin as an env var and set it to `http://localhost:8081` in DevelopmentConfiguration.

(@dchiquito  is that right?)

zachmullen (changes requested):

miqa/urls.py
annehaley: we could put something like `HOMEPAGE = HomePageView.as_view()` in MiqaMixin and `HOMEPAGE = RedirectView.as_view(url='http://localhost:8081')` in DevelopmentConfiguration?

zachmullen: Sure, that sounds good. 👍

miqa/urls.py
```diff
@@ -36,8 +39,9 @@
     permission_classes=(permissions.IsAuthenticated,),
 )
 
+
 urlpatterns = [
-    path('', HomePageView.as_view(), name='home'),
+    settings.HOMEPAGE,
```
dchiquito, line 44:
```suggestion
    path('', settings.HOMEPAGE_VIEW, name='home')
```
I think that's a little more DRY, and makes it a little clear that you're shoving a view into a setting. It also simplifies the properties a bit.

dchiquito (changes requested):

miqa/urls.py
annehaley: I'm getting a strange error when I separate them like that:
`TypeError: Couldn't setup configuration 'miqa.settings.DevelopmentConfiguration':  view() missing 1 required positional argument: 'request'`.

I get this when urlpatterns contains `path('', sesttings.HOMEPAGE_VIEW, name='home)` and settings.HOMEPAGE_VIEW returns `RedirectView.as_view(url='https://localhost:8081')`. It's as if having them separate breaks how `as_view` works with `path` to receive a request?

miqa/urls.py
annehaley: I did run into something for this. I noticed that this works fine for the DevelopmentConfiguration, where all we need is RedirectView imported. However, for all other configurations inheriting from MiqaMixin, we need to import HomePageView from `miqa.core.rest.home`. When such an import is made anywhere in settings.py, we get this error:
`django.core.exceptions.AppRegistryNotReady: Couldn't setup configuration 'miqa.settings.DevelopmentConfiguration':  Apps aren't loaded yet.`

I couldn't find a workaround that would allow us to do lazy-loading on this import (even though this import doesn't even reference any models). Thankfully, the functionality of HomePageView is really simple, so what we _could_ do is use TemplateView directly: `return path('', TemplateView.as_view(template_name='index.html'), name='home')`. In this case, we wouldn't even need the file miqa/core/rest/home.py, though deleting it would lose the comment contents...

What do you think, @zachmullen?

dchiquito (approved):
LGTM

zachmullen (approved):

miqa/urls.py
annehaley: @dchiquito and I were able to figure out a new solution where the settings configurations can either specify a string URL or `None` for a value called `HOMEPAGE_REDIRECT_URL`. home.py returns the original `TemplateView` if this value is `None` or a `RedirectView` to the URL otherwise. We figured that this solution is both DRY and Configuration-flexible.

Issue #117, opened 2021-09-01
Unable to import sessions from CSV


jimklo:
When trying to import sessions from a CSV file; we receive an error in the log and no sessions are actually imported:

```
django_1    | Django version 3.2.6, using settings 'miqa.settings'
django_1    | Starting development server at http://0.0.0.0:8000/
django_1    | Quit the server with CONTROL-C.
django_1    | [22:09:10] INFO     "GET / HTTP/1.1" 200 1152                    basehttp.py:161
django_1    | [22:09:11] INFO     "GET /api/v1/users/me HTTP/1.1" 200 83       basehttp.py:161
django_1    |            INFO     "GET /api/v1/sessions HTTP/1.1" 200 166      basehttp.py:161
django_1    |            INFO     "GET /api/v1/sessions/900a4860-cbae-44c9-a4f basehttp.py:161
django_1    |                     f-46fc326aa55a HTTP/1.1" 200 690
django_1    |            INFO     "GET /api/v1/sites HTTP/1.1" 200 230         basehttp.py:161
django_1    | field names: ['xnat_experiment_id', 'nifti_folder', 'scan_id', 'scan_type', 'experiment_note', 'decision', 'scan_note']
django_1    | common path prefix: /fs/storage/xnat-kitware/archive/sri_incoming/arc001/B-00001-F-2-20160127/RESOURCES/
django_1    | [22:09:17] ERROR    Internal Server Error: /miqa4/api/v1/sessions/900 log.py:224
django_1    |                     a4860-cbae-44c9-a4ff-46fc326aa55a/import
django_1    |                     Traceback (most recent call last):
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/core/handlers/exception.py", line 47, in
django_1    |                     inner
django_1    |                         response = get_response(request)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/core/handlers/base.py", line 181, in
django_1    |                     _get_response
django_1    |                         response = wrapped_callback(request,
django_1    |                     *callback_args, **callback_kwargs)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/dj
django_1    |                     ango/views/decorators/csrf.py", line 54, in
django_1    |                     wrapped_view
django_1    |                         return view_func(*args, **kwargs)
django_1    |                       File "/usr/local/lib/python3.8/site-packages/re
[...]

PR #137, opened 2021-09-02
Better import errors

dchiquito:
Jim's problems in #117 are utterly inscrutable because we attempt to wrap any errors in another exception, which obscures the stack trace. By removing the error handling entirely, we let django throw a 500 error on failed imports and print the stack trace.

Also add some meaningful tests for import.

### DIFF
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -16,20 +16,14 @@ def import_data(user, session: Session):
     if session.import_path.endswith('.csv'):
         with open(session.import_path) as fd:
             csv_content = fd.read()
-            try:
-                json_content = csvContentToJsonObject(csv_content)
-                validate(json_content, schema)  # TODO this should be an internal error
-            except (ValidationError, Exception) as e:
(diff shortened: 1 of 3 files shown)


curtislisle:
One of the last things Scott Wittenburg added to the previous codebase (girder3) was to add nice support for import errors.  I don't think this ever made it into master.  Here is his commit on the previous codebase: https://github.com/scottwittenburg/miqa/commit/57fdc5fee5dff86a01539a1055d3442637ff1785

I built it locally and it popped up a dialog explaining what was wrong with the JSON or CSV import, if I recall.   Maybe some of this is still useful, or maybe it is mute, since Daniel is cranking on this problem already.   Just figured I'd toss this comment on the table in case it is helpful

zachmullen (approved):
Thanks for adding tests.

dchiquito:
@curtislisle AFAICT his changes are already present in our current code base. Most of that change seems to be around the frontend, so we already have that messaging in the UI.

PR #140, opened 2021-09-07
Typescript Conversions #2

annehaley:
Converted six more files to use typescript, starting with the files that had the fewest errors. Added @types file to define User and Session. Finally, made some small changes to the eslintrc file so that typescript could be correctly interpreted.

### DIFF
--- a/client/src/components/EmailRecipientCombobox.vue
+++ b/client/src/components/EmailRecipientCombobox.vue
@@ -1,4 +1,5 @@
-<script>
+<script lang="ts">
+import { PropType } from 'vue';
 export default {
   name: 'EmailRecipientCombobox',
   props: {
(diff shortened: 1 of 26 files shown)


annehaley:
I'm not sure why test-client isn't passing; running tox locally gets a perfect score and I can't figure out the meaning of the error trace here.

client/src/components/EmailRecipientCombobox.vue
```diff
@@ -11,7 +13,7 @@ export default {
       type: Array,
     },
     candidates: {
-      type: Array,
+      type: Array as PropType<Array<string>>,
```
dchiquito, line 16:
```suggestion
      type: Array as PropType<string[]>,
```
I think `type[]` is generally preferred to `Array<type>`, although I personally do not feel strongly on the matter.

dchiquito (approved):
The actual changes here are pretty minimal, it's mostly autofixing tag argument ordering.

PR #141, opened 2021-09-08
Lock UI changes

### LINKED ISSUE #118: Move lock to experiment line
Currently the lock is on the bottom:

image

The idea is to move the icon to the experiment line so it's clear it's associated with the whole experiment.

### LINKED ISSUE #119: Consider reversing metaphor so unlock means you can edit
There is an existing metaphor that when apps show things as locked you cannot edit, while the app does the reverse. For example on mac, this indicates you cannot edit (see bottom):

image

Consider using a locked icon if you cannot edit it, and convert to unlocked when you can edit.

annehaley:
Resolves #118, resolves #119

### DIFF
--- a/client/src/components/ExperimentLockIcon.vue
+++ b/client/src/components/ExperimentLockIcon.vue
@@ -31,17 +31,17 @@ export default {
         icon
         v-bind="attrs"
         v-on="on"
-        @click="lockExperiment(experiment)"
+        @click="unlockExperiment(experiment)"
       >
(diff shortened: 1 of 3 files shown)


dchiquito (approved):
These changes look good to me, assuming the UX team agrees this is a good layout.

annehaley:
Choosing to abandon this PR in favor of the new UI redesign effort, which is a much more thorough interface revision.

PR #115, opened 2021-09-01
Refactor the NN to produce multiple outputs

dzenanz:
Closes #30.

PR #144, opened 2021-09-14
NN integration 1

annehaley:
This PR is meant to demonstrate a proof-of-concept on invoking the NN evaluation on scans that are imported. Right now, when the user imports scans, an evaluation of those scans is run synchronously and the result is simply printed to the django terminal. 

Next PRs will be for creating django models for Evaluations and NN_models, then for making the evaluation runs asynchronous, then for incorporating Evaluations into the UI.

For reviewing this request, I ask that you try out the import please. I got it working in my environment, but I'd like to know how well that works on other machines, because docker needs to use the GPU for the evaluation.


annehaley:
I now see that one CI test is failing; from the logs, it is getting an Unpickling Error for the import test. I have found that this occurs when the NN attempts to invoke a model file that has not yet been retrieved from Git LFS. Right now, the repo contains pointer files instead of the actual content.

Could I get some help adjusting the CI VM so that Git LFS can work correctly when these tests are run? On my local machine, I had to do the following: 

```
sudo apt install git-lfs
git lfs install
git lfs fetch
git lfs checkout
```

These will convert the lfs pointer files to their actual content so that the NN can actually obtain the content of the model files @dzenanz uploaded to the repo. Is this possible to do for the CI tests?

miqa/learning/nn_classifier.py
```diff
@@ -326,10 +342,10 @@ def forward(self, output, target):
         return loss
 
 
-def convert_bool_to_int(value):
-    if value == True:
+def convert_bool_to_int(value: bool):
+    if value:
         return 1
-    elif value == False:
+    elif not value:
```
zachmullen, line 348: This entire function scares me... but it's worth mentioning that one of  `if value` and `if not value` will always be true, hence the below `else` block would never be entered. It seems like maybe this supports passing in types besides `bool`, but I have no idea how it's being used. For now, I'd suggest leaving the old behavior. If this was to fix a linting issue, maybe we should just use `is False` rather than `== False`.

dzenanz:
Does adding `git-lfs` to the [list of packages](https://github.com/OpenImaging/miqa/blob/e5eeb9574e8750a8f0f4dc4eeee9b60122c4d6a9/Dockerfile#L6) help? We probably need to also add `git lfs install` etc either before or after [install tox](https://github.com/OpenImaging/miqa/blob/e5eeb9574e8750a8f0f4dc4eeee9b60122c4d6a9/.github/workflows/ci.yml#L38-L41).

prod/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
       POSTGRES_DB: django
       POSTGRES_PASSWORD: ${DOCKER_POSTGRES_PASSWORD}
     ports:
-      - ${DOCKER_POSTGRES_PORT-5432}:5432
+      - ${DOCKER_POSTGRES_PORT-5432}
```
zachmullen, line 9: Is this change related to the PR topic? If not, can we isolate it to a separate PR to capture its intent?

prod/docker-compose.yml
```diff
@@ -28,8 +28,17 @@ services:
...
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: 1
+              capabilities: [gpu]
     volumes:
       - "${MIQA_MOUNT_DIR}:${MIQA_MOUNT_DIR}"
+      - "/home/anne/miqa/samples:/mnt/mydata"
+      - "/home/anne/miqa/samples:/srv/samples"
```
zachmullen, line 41: These should be relative paths

miqa/learning/nn_classifier.py
```diff
@@ -282,9 +287,20 @@ def evaluate1(model_path, image_path):
...
+    for index, artifact_name in enumerate(artifacts):
+        labeled_results[artifact_name] = result[index + 3]
+    return labeled_results
+
 
 class CombinedLoss(torch.nn.Module):
-    def __init__(self, binary_class_weights, focal_loss=monai.losses.FocalLoss()):
+    default_focal_loss = monai.losses.FocalLoss()
+
+    def __init__(self, binary_class_weights, focal_loss=default_focal_loss):
```
zachmullen, line 303: In Python, [we should avoid mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

```python
def __init__(self, binary_class_weights, focal_loss=None):
    if focal_loss is None:
        focal_loss = monai.losses.FocalLoss()

**Review by zachmullen — CHANGES_REQUESTED** [review]

File: .github/workflows/ci.yml [hunk 5]
```diff
@@ -38,6 +38,9 @@ jobs:
       - name: Install tox
         run: |
           pip install --upgrade pip
+          git lfs install
+          git lfs fetch
+          git lfs checkout
```
zachmullen, line 43: In other projects that use LFS in CI, I see the following changes:

```yml
    steps:
      - uses: actions/checkout@v2
        with:
          lfs: true
```

I think that is enough, and then you can omit these run steps.

prod/docker-compose.yml
annehaley: This was actually done when I was setting up miqademo, and it's purely related to prod setup. We can separate it into another PR, good catch.

prod/docker-compose.yml
annehaley: I meant to leave this whole file out of the commit, whoops. I'll fix it.

learning/README.md
```diff
@@ -32,10 +32,30 @@ python ./miqa/learning/nn_classifier.py -f ./T1_fold -c 3 --all
...
+```
+
+This will convert your pointer files. You should now be able to use the files in the models subdirectory.
+
+
 ## Run inference
 To run inference on a single file, execute:
 ```shell
-python ./miqa/learning/nn_classifier.py -m ./miqa01-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
+python ./miqa/learning/nn_classifier.py -m ./models/miqa01-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
```
dzenanz, line 58: `miqa01` should be renamed into `miqaT1`. I forgot to do this as part of #115.

miqa/learning/nn_classifier.py
annehaley: It was to fix a linting error. I changed it to use `is` instead of `==`.

prod/docker-compose.yml
annehaley: I put the `:5432` back in, and we can look at this change later on another PR.

prod/docker-compose.yml
annehaley: I realize I had committed it because of the nvidia driver change, so I reconciled that change with removing my local-fix changes. Now this file does not change the postgres port or the django mounted volumes, but does change the GPU resource for django.

.github/workflows/ci.yml
annehaley: this change successfully converts the models from pointer files to their true content. Made this fix and the one comment below.

annehaley:
Now the CI test is failing for a different reason; instead of failing on using the model file, it fails to use the GPU. Do either of you know how to have the CI test use an Nvidia GPU? @zachmullen @dzenanz

zachmullen:
We won't be able to have GPUs in our CI environment, and also we'll want to support CPU mode in dev as well.

@dzenanz what needs to change in your script to fall back to CPU when cuda is not available?

dzenanz (commented):
Code looks good. I will try it too.

dzenanz:
@zachmullen Probably [this](https://pytorch.org/tutorials/recipes/recipes/save_load_across_devices.html):

When loading a model on a CPU that was trained with a GPU, pass torch.device('cpu') to the map_location argument in the torch.load() function.
```py
# Specify a path to save to
PATH = "model.pt"

# Save
torch.save(net.state_dict(), PATH)

# Load
device = torch.device('cpu')
model = Net()
model.load_state_dict(torch.load(PATH, map_location=device))
```

dzenanz:
I think we only need to add `map_location=device` parameter to all `torch.load()` calls, of which there are only a few.

annehaley:
@dzenanz what should be the determinant of what is assigned to `device`? should there be an if statement or a try-catch situation?

dzenanz:
In a few places there is this: `device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')`.

annehaley:
that's perfect, I'll make that change.

dzenanz:
So there is already a variable called `device` before `torch.load()` is invoked.

learning/README.md
annehaley: made this change in the README, the files were already renamed.

annehaley:
great! thanks to you both for your help, the ci tests are passing!

zachmullen (approved):
If you're looking to merge this and continue on future PR's, that's fine. I will open an issue about sharing the model loading across image runs (i.e. not using `evaluate1`).

annehaley:
Yes, that's the plan. I'll move forward with the model definitions first, because @dzenan wanted to work on refactoring `nn_classifier.py` today. To avoid needing to resolve conflicts, I'll work keep from editing that file until he's done. Then I'll be able to add another function that satisfies this.

dzenanz:
I would like to try this on my system before merging. I also think we should squash-merge as there are a lot of back-and-forth commits. I will "approve" if this runs on my computer. Are readmes for the rest of this repository up to date?

dzenanz:
I will get to this late today or tomorrow morning.

dzenanz:
The evaluations are printed to console on step 5 `docker-compose run --rm django ./manage.py populate --csv /srv/samples/scans_to_review-2019-01-23.csv`.

dzenanz (approved):
Good to merge. Please squash-merge - it can be done using GitHub's UI.

dzenanz:
Manually squash-merged as de0905c632b377a8bbee32de50fbe8875402fbe7.

PR #147, opened 2021-09-15
Update getting started documentation

dzenanz:
Inspired by following the instructions in order to reproduce #144.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,10 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker-compose run --rm npm npm install`
+1. Run `docker-compose run --rm npm npm install` (`docker-compose` must be at least version 1.28)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,12 +4,10 @@
 This is the simplest configuration for developers to start with.
 
 ### Initial Setup
-1. Run `docker-compose run --rm npm npm install`
+1. Run `docker-compose run --rm npm npm install` (`docker-compose` must be at least version 1.28)
```
annehaley, line 7: I know that I was the one to write this line in, but after using these instructions a few times, we should clarify that (1) should be run in the `client/` directory, too. Perhaps that means doing the following:

```
1. cd client
2. docker-compose run --rm npm npm install
3. cd ..
```

dzenanz: I ran this command from repository root, and things worked. So I think `cd client` is not necessary.

annehaley: Fair enough, comment revoked :)

PR #148, opened 2021-09-15
NN integration 2

annehaley:
This PR includes models for NNModel and Evaluation and demonstrates that Evaluation objects can be created upon the ingest of new data.

Objects of the type NNModel represent trained models; this django model is a wrapper around trained .pth files. This model, unlike others in MIQA, should come pre-populated. There is a new file that allows us to specify the trained models that exist in learning/models so that the database will create objects for them if they do not already exist. This django model is necessary for a user to specify which NNModel they would like to use. The Session object now includes a field to point to the selected NNModel.

Objects of the type Evaluation are created for Images when they are imported. Evaluation objects include the results of running the associated NNModel, and must have keys that exactly match the outputs specified on that NNModel. These Evaluation objects will be what we use to display the results of the NN evaluation to the user.

The next PR will be to make the Evaluation process asynchronous.


annehaley:
Not sure why commits from the previous PR are shown here, but the differences start at commit 
ad63b28.

annehaley:
I thought it might help to include a concept document that was made during brainstorming:

NN Integration classes drawio

dzenanz:
Rebasing on current master should get rid of the old commits.

We should allow computing image quality metrics (IQMs) using MRIQC. This will not have a `.pth` file. Perhaps `NNModel` should be renamed (or generalized) to `ImageMetricsModel` to plug MRIQC there?

annehaley:
I rebased and the number of commits at the top went down, but this PR still lists them all as separate commits.

As for generalizing NNModel, that depends on how the user would expect to interact with the NN evaluations and the MRIQC evaluations. Would the user choose one and only get that kind of evaluation for the whole session?  Or would they expect to get results from more than one? One of each?

dzenanz:
I guess we could allow user to check the models they want to be used for evaluations. Make all checked by default?

I now rebased only your new commits and force-pushed.

annehaley:
Thanks. So perhaps a Session should point to one or more of these objects, which could be generalized as "EvaluationModel"s? Having them all checked by default may cause a lot of extra computation, though, which could be unnecessary for the average user who just wants some evaluation information. We would need to determine the real differences between these models to say which should be checked by default and which are extra features.

dzenanz:
Sure, we could only have the computationally cheap ones "on by default".

dzenanz (commented):
I glanced the code, but I am not familiar with django so I will leave review to others.

annehaley:
One more question about smushing together the ideas of MRIQC and your own NN models: would they be invoked in the same manner? Or would we need some if-else logic on whether the `model_file` is referencing a .pth file or not? This might be a reason to keep them as separate django models.

curtislisle:
> One more question about smushing together the ideas of MRIQC and your own NN models: would they be invoked in the same manner? Or would we need some if-else logic on whether the `model_file` is referencing a .pth file or not? This might be a reason to keep them as separate django models.

I would say that NNModel is a proper subclass of EvaluationModel, which happens to need a model_file reference.  We will want to be able to load a few models in MIQA, and some (but not all) of the models will be need to initialized by reading their model file data.  All model files might not have the .pth extension, also.   For example, it is possible to use Fast.ai (another framework that uses PyTorch) and have pickle files for models instead.    We don't have to design for every eventuality, but I wanted us to at least think about this and made rational tradeoffs.  Thanks for your thinking and discussion about this issue.

annehaley:
> I would say that NNModel is a proper subclass of EvaluationModel, which happens to need a model_file reference. We will want to be able to load a few models in MIQA, and some (but not all) of the models will be need to initialized by reading their model file data. All model files might not have the .pth extension, also. For example, it is possible to use Fast.ai (another framework that uses PyTorch) and have pickle files for models instead. We don't have to design for every eventuality, but this design time is the right time to at least think about this and make rational tradeoffs. Thanks for your designs and discussion about this issue.

That sounds like a good idea, to have some inheritance structure of the Django models. It will allow us to combine shared functionality and specify non-shared functionality between the different types of EvaluationModels.

annehaley:
This PR is old and we changed our concept of the NNEvaluation model. Making a new PR to keep it tidy.

PR #150, opened 2021-09-16
Refactor NN classifier

dzenanz:


annehaley (changes requested):
So far the refactoring looks good, but it may be helpful to split this large file in two. Perhaps separate the training functions from the evaluation functions?

dzenanz:
My first though was that splitting into two files would make it more complicated, or would require some duplication.

On second thought, it would make sense to put evaluation into its own file with less imports, so startup time would be shorter. I now plan to do that at a later date, unless you want to do this split within the next few days?

annehaley:
Sounds good. Then the only thing to do before merge would be getting the ci test to pass with some style changes. Splitting in two can be the next PR.

dzenanz:
How to ignore `miqa\learning\wandb` folder? Tox spews out tons of errors for old versions of the script:
```text
...
miqa\learning\wandb\run-20210831_122216-2yg825x8\files\code\learning\nn_classifier.py:337:14: E712 comparison to True should be 'if cond is True:' or 'if cond:'
    if value == True:
             ^
miqa\learning\wandb\run-20210831_122216-2yg825x8\files\code\learning\nn_classifier.py:339:16: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
    elif value == False:
               ^
miqa\learning\wandb\run-20210831_122216-2yg825x8\files\code\learning\nn_classifier.py:359:20: B007 Loop control variable 'artifact' not used within the loop body. If this is intended, start the name with an underscore.
            for i, artifact in enumerate(artifacts):
                   ^
miqa\learning\wandb\run-20210906_104638-4olf182f\files\code\learning\nn_classifier.py:18:1: F401 'torchio.transforms' imported but unused
from torchio import transforms
^
miqa\learning\wandb\run-20210906_104638-4olf182f\files\code\learning\nn_classifier.py:293:57: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
    def __init__(self, binary_class_weights, focal_loss=monai.losses.FocalLoss()):
                                                        ^
miqa\learning\wandb\run-20210906_104638-4olf182f\files\code\learning\nn_classifier.py:336:14: E712 comparison to True should be 'if cond is True:' or 'if cond:'
    if value == True:
             ^
miqa\learning\wandb\run-20210906_104638-4olf182f\files\code\learning\nn_classifier.py:338:16: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
    elif value == False:
               ^
miqa\learning\wandb\run-20210906_104638-4olf182f\files\code\learning\nn_classifier.py:358:20: B007 Loop control variable 'artifact' not used within the loop body. If this is intended, start the name with an underscore.
            for i, artifact in enumerate(artifacts):
                   ^
miqa\learning\wandb\run-20210916_143416-2i16ruop\files\code\miqa\learning\nn_classifier.py:488:43: F821 undefined name 'device'
        evaluate_model(model, val_loader, device, writer, 0, 'val')
                                          ^
miqa\learning\wandb\run-20210916_143416-2i16ruop\files\code\miqa\learning\nn_classifier.py:490:45: F821 undefined name 'device'
        evaluate_model(model, train_loader, device, writer, 0, 'train')
                                            ^
miqa\learning\wandb\run-20210916_143416-2i16ruop\files\code\miqa\learning\nn_classifier.py:491:16: F821 undefined name 'sizes'
        return sizes
               ^
miqa\learning\wandb\run-20210916_143416-2i16ruop\files\code\miqa\learning\nn_classifier.py:563:12: F821 undefined name 'sizes'
    return sizes
           ^
```

annehaley:
If the wandb directory is generated for training, you could leave it out of the repo entirely and include it in the gitignore. I don't remember it containing any original code/functionality.

dzenanz:
It doesn't contain any original code, but it contains old versions of the script which used it, one for each saved run.

@zachmullen @dchiquito I went through a few different suggestions from [SO](https://stackoverflow.com/questions/11117062/how-to-tell-py-test-to-skip-certain-directories) but none worked. Do you know how to make tox ignore `wandb` directory?

PR #151, opened 2021-09-17
Fix types

dchiquito:
The frontend was pretty broken, I repaired everything I could find.

@annehaley would you mind taking this for a test drive and hunting for any broken features?


dchiquito:
I think the Scan -> Settings -> Scan problem has basically always been there, since it was a gritty issue with the Vuex store being poorly implemented. Since I tracked it down though I fixed it here, great catch!

PR #152, opened 2021-09-20
NN integration 2

annehaley:
This is the new PR for adding the Evaluation django model and the EvaluationModel static python class. The task `evaluate_images` in `tasks.py` now uses both of these objects. Each image in an import gets an Evaluation object, and the EvaluationModels are only loaded once into memory (even if multiple images in that import will use that model).


miqa/learning/README.md
```diff
@@ -55,7 +55,7 @@ This will convert your pointer files. You should now be able to use the files in
 ## Run inference
 To run inference on a single file, execute:
 ```shell
-python ./miqa/learning/nn_classifier.py -m ./models/miqaT1-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
+python ./miqa/learning/nn_classifier.py -m ./models/miqa01-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
```
dzenanz: This seems like a rebase gone wrong, only reverting this recent change of file names.

miqa/learning/nn_classifier.py
```diff
@@ -288,9 +288,9 @@ def evaluate1(model_path, image_path):
     print(f'Overall quality of {image_path}, on 0-10 scale: {result[0]:.1f}')
 
     labeled_results = {
-        'overall_quality': result[0],
-        'signal_noise_ratio': result[1],
-        'contrast_noise_ratio': result[2],
+        "overall_quality": result[0],
```
dzenanz: Another rebase gone wrong only changing the quote style?

miqa/learning/README.md
```diff
@@ -55,7 +55,7 @@ This will convert your pointer files. You should now be able to use the files in
 ## Run inference
 To run inference on a single file, execute:
 ```shell
-python ./miqa/learning/nn_classifier.py -m ./models/miqa01-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
+python ./miqa/learning/nn_classifier.py -m ./models/miqaT1-val0.pth -1 ./PredictHD_small/sub-699312/ses-38845/anat/sub-699312_ses-38845_run-003_BADT1w.nii.gz
```
dzenanz: All the commits until this one (including this one) have been integrated by de0905c632b377a8bbee32de50fbe8875402fbe7. They have been squashed into a single commit, that is probably what is confusing you. You can (and should) skip this and previous commits on this branch.

miqa/core/migrations/0013_auto_20210917_1235.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.7 on 2021-09-17 12:35
```
dzenanz, line 1: If this file is auto-generated, should we be versioning it?

If it is not auto-generated, shouldn't it be given a meaningful name?

miqa/core/models/scan.py
```diff
@@ -5,6 +5,8 @@
 from django.db import models
 from django_extensions.db.models import TimeStampedModel
 
+SCAN_TYPES = [('ncanda-t1spgr-v1', 'ncanda-t1spgr-v1'), ('ncanda-mprage-v1', 'ncanda-mprage-v1')]
```
dzenanz, line 8: `ncanda-mprage-v1` is repeated twice. Is this intentional?

miqa/learning/nn_classifier.py
```diff
@@ -260,7 +260,7 @@ def evaluate_model(model, data_loader, device, writer, epoch, run_name):
             return outputs
 
 
-def evaluate1(model_path, image_path):
+def evaluate1(model, image_path):
```
dzenanz, line 263: The invocation in "main" function should be changed too. It currently reads `evaluate1(args.modelfile, args.evaluate1)`

miqa/core/migrations/0014_evaluation.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.7 on 2021-09-20 16:55
```
dzenanz, line 1: Should we version auto-generated files? Should we give them a better name?

miqa/core/models/evaluation.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __str__(self):
+        return f'Evaluation for {str(self.image.raw_path)}'
+
+    def save(self, *args, **kwargs):
+        evaluation_model = available_evaluation_models[self.evaluation_model]
+        if any([key not in evaluation_model.expected_outputs for key in self.results.keys()]):
+            raise ValidationError(
+                'Result keys must match expected outputs of associated evaluation model.'
+            )
```
zachmullen, line 23: Let's move validation logic into `full_clean` rather than overriding `save`. Then we'll need to call that in places where we make the model (at least if we're using user-provided input to these fields).

miqa/core/models/evaluation.py
```diff
@@ -0,0 +1,25 @@
...
+    image = models.ForeignKey('Image', null=False, on_delete=models.CASCADE)
+    evaluation_model = models.CharField(max_length=50)
+    results = models.JSONField()
+
+    def __str__(self):
+        return f'Evaluation for {str(self.image.raw_path)}'
+
+    def save(self, *args, **kwargs):
+        evaluation_model = available_evaluation_models[self.evaluation_model]
+        if any([key not in evaluation_model.expected_outputs for key in self.results.keys()]):
```
zachmullen, line 20: Any is one of the special built-in functions that have special syntax we can take advantage of:

```suggestion
        if any(key not in evaluation_model.expected_outputs for key in self.results.keys()):
```

miqa/core/models/scan.py
annehaley: Yes, when specifying `choices` for a django `CharField`, the items in the list must be tuples of `(label, value)`. In many cases, it is good practice to have the label and value the same.

miqa/learning/nn_classifier.py
annehaley: Good catch. Updated.

zachmullen (changes requested):

miqa/core/models/evaluation.py
annehaley: We actually don't use any user input when creating Evaluation objects. This is just a sanity check. Should we move it to `full_clean` still?

zachmullen: Oh, in that case I would actually just delete this code. My general philosophy here is that we shouldn't include code to try to protect us against bugs of our own making, unless there's some compelling reason to believe we would cause a specific bug (which is maybe a design red flag that could be addressed with refactoring).

annehaley: Okay, removing the check. The only place we would create Evaluation objects is in the one place it is currently called, in `evaluate_images` in `tasks.py`.

miqa/core/migrations/0014_evaluation.py
annehaley: Migration files are generated by Django to prep the db tables. My attempts to rename the migration files have only given me errors so far. I'm sure it's possible but not sure how to accomplish it. @zachmullen do you know how?

zachmullen: It's common to rename these files, however you'll have to rename them _before_ you apply the migration IIRC. You could also migrate down, rename it, then migrate back up.

miqa/core/models/session.py
```diff
@@ -12,6 +20,28 @@ class Session(TimeStampedModel, models.Model):
...
+        if any(
+            value is None or value not in available_evaluation_models
+            for value in self.evaluation_models.values()
+        ):
+            raise ValidationError(
+                f'Values in evaluation models must be valid evalution model names. '
+                f'Valid evaluation model names are {available_evaluation_models.keys()}'
+            )
+
+        super(Session, self).clean()
```
zachmullen, line 47: Since we are python 3.8+, you can just do `super().clean()`

PR #153, opened 2021-09-21
Do not create an extra worker process in evaluate1

dzenanz:
To avoid `Task miqa.core.tasks.evaluate_data[4c98627b-d341-47ea-9a6b-a7836d515f00] raised unexpected: AssertionError('daemonic processes are not allowed to have children')`

### DIFF
--- a/miqa/learning/nn_classifier.py
+++ b/miqa/learning/nn_classifier.py
@@ -284,7 +284,7 @@ def evaluate1(model, image_path):
         transform=train_transforms,
     )
     evaluation_loader = DataLoader(
-        evaluation_ds, batch_size=1, num_workers=1, pin_memory=torch.cuda.is_available()
+        evaluation_ds, batch_size=1, pin_memory=torch.cuda.is_available()
     )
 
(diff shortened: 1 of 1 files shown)


miqa/learning/nn_classifier.py
```diff
@@ -284,7 +284,7 @@ def evaluate1(model, image_path):
         transform=train_transforms,
     )
     evaluation_loader = DataLoader(
-        evaluation_ds, batch_size=1, num_workers=1, pin_memory=torch.cuda.is_available()
+        evaluation_ds, batch_size=1, num_workers=0, pin_memory=torch.cuda.is_available()
```
zachmullen, line 287: I think I'd prefer if we just omit the kwarg altogether; seems less confusing in terms of intent.

zachmullen (changes requested):

zachmullen (approved):

dzenanz:
I had those explicit to ease changes during experimentation. I guess we will always want 0 workers there.

dzenanz:
@annehaley merge if happy and CI passes.

annehaley (approved):

PR #154, opened 2021-09-21
NN integration 3

### LINKED ISSUE #126: NN integration: DevOps to get models built into images

### LINKED ISSUE #127: NN integration: Django models for NN outputs, linked to scan
The NN is run on a scan, not on an image.

Outputs from the NN:
* a grade [0,10], which indicates the overall quality of the scan. 
* 10 classification probabilities [0,1]. These should be rendered in the UI as tags if they are above a cutoff.

### LINKED ISSUE #128: NN integration: celery task that runs on ingest

annehaley:
Changes the invocation of Celery tasks. Previously, these tasks were invoked synchronously, never getting sent to Celery. Now they are called asynchronously and the import task finishes before spawning a single background task to evaluate all the images in the import. Evaluation objects are created in this background task.

Resolves #126. Resolves #127. Resolves #128.

### DIFF
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -2,11 +2,13 @@ import json
 import os
 from pathlib import Path
 import re
-from typing import List
+
+from django.contrib.auth.models import User
 
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 
+from miqa.celery import app
 from miqa.core.conversion.csv_to_json import csvContentToJsonObject, find_common_prefix
 from miqa.core.conversion.json_to_csv import jsonObjectToCsvContent
 from miqa.core.models import (
@@ -25,7 +27,14 @@ from miqa.learning.evaluation_models import available_evaluation_models
 from miqa.learning.nn_classifier import evaluate1
 
 
-def evaluate_data(images: List[Image], session: Session):
+@app.task
+def evaluate_data(image_ids, session_id):
+    try:
+        images = [Image.objects.get(id=image_id) for image_id in image_ids]
+        session = Session.objects.get(id=session_id)
+    except (Image.DoesNotExist, Session.DoesNotExist):
+        raise ValueError('Provide the id of the user and the id of the session.')
+
     loaded_evaluation_models = {}
     for image in images:
         scan_type = [image.scan.scan_type][0]
@@ -45,7 +54,14 @@ def evaluate_data(images: List[Image], session: Session):
         evaluation.save()
 
 
-def import_data(user, session: Session):
+@app.task
+def import_data(user_id, session_id):
+    try:
+        user = User.objects.get(id=user_id)
+        session = Session.objects.get(id=session_id)
+    except (User.DoesNotExist, Session.DoesNotExist):
+        raise ValueError('Provide the id of the user and the id of the session.')
+
     if session.import_path.endswith('.csv'):
         with open(session.import_path) as fd:
             csv_content = fd.read()
@@ -126,10 +142,17 @@ def import_data(user, session: Session):
     ScanNote.objects.bulk_create(notes)
     Annotation.objects.bulk_create(annotations)
 
-    evaluate_data(images, session)
+    evaluate_data.delay([image.id for image in images], session.id)
 
 
-def export_data(user, session: Session):
+@app.task
+def export_data(user_id, session_id):
+    try:
+        user = User.objects.get(id=user_id)
+        session = Session.objects.get(id=session_id)
+    except (User.DoesNotExist, Session.DoesNotExist):
+        raise ValueError('Provide the id of the user and the id of the session.')
+
     data_root = None
     experiments = []
     scans = []
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -23,7 +23,7 @@ from miqa.core.models import (
 )
 from miqa.core.schema.data_import import schema
 from miqa.learning.evaluation_models import available_evaluation_models
-from miqa.learning.nn_classifier import evaluate1
+from miqa.learning.nn_classifier import evaluate_many
 
 
 @shared_task
@@ -31,23 +31,27 @@ def evaluate_data(image_ids, session_id):
     images = Image.objects.filter(pk__in=image_ids)
     session = Session.objects.get(id=session_id)
 
-    loaded_evaluation_models = {}
+    model_to_images_map = {}
     for image in images:
-        scan_type = [image.scan.scan_type][0]
-        eval_model_name = session.evaluation_models[scan_type]
-        # only load the necessary models into memory once each
-        if eval_model_name not in loaded_evaluation_models:
-            loaded_evaluation_models[eval_model_name] = available_evaluation_models[
-                eval_model_name
-            ].load()
-        # perform evaluation on each
-        current_model = loaded_evaluation_models[eval_model_name]
-        evaluation = Evaluation(
-            image=image,
-            evaluation_model=eval_model_name,
-            results=evaluate1(current_model, str(image.raw_path)),
+        eval_model_name = session.evaluation_models[[image.scan.scan_type][0]]
+        if eval_model_name not in model_to_images_map:
+            model_to_images_map[eval_model_name] = []
+        model_to_images_map[eval_model_name].append(image)
+
+    for model_name, image_set in model_to_images_map.items():
+        current_model = available_evaluation_models[model_name].load()
+        results = evaluate_many(current_model, [str(image.raw_path) for image in image_set])
+
+        Evaluation.objects.bulk_create(
+            [
+                Evaluation(
+                    image=image,
+                    evaluation_model=model_name,
+                    results=results[str(image.raw_path)],
+                )
+                for image in image_set
+            ]
         )
-        evaluation.save()
 
 
 @shared_task
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -2,8 +2,8 @@ import json
 import os
 from pathlib import Path
 import re
-from celery import shared_task
 
+from celery import shared_task
 from django.contrib.auth.models import User
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -2,12 +2,12 @@ import json
 import os
 from pathlib import Path
 import re
+from celery import shared_task
 
 from django.contrib.auth.models import User
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 
-from miqa.celery import app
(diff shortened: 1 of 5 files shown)


miqa/core/tasks.py
```diff
@@ -25,7 +26,14 @@
 from miqa.learning.nn_classifier import evaluate1
 
 
-def evaluate_data(images: List[Image], session: Session):
+@app.task
+def evaluate_data(image_ids, session_id):
+    try:
+        images = [Image.objects.get(id=image_id) for image_id in image_ids]
```
zachmullen, line 32: We can retrieve these with a single query rather than n queries using `filter(pk__in=image_ids)`

miqa/core/tasks.py
```diff
@@ -25,7 +26,14 @@
...
+        session = Session.objects.get(id=session_id)
+    except (Image.DoesNotExist, Session.DoesNotExist):
+        raise ValueError('Provide the id of the user and the id of the session.')
```
zachmullen, line 35: We can remove this `try/except`; the caller should be responsible for validating the IDs being passed in. And if for some reason it doesn't exist, it's preferable just to bubble up the relevant `DoesNotExist` exception.

Also, as a general pattern, if you have two statements that each trigger a certain type of exception, it would be preferable to put them in two separate try/except blocks, one for each of the statement/exception type pairs.

miqa/core/tasks.py
```diff
@@ -25,7 +26,14 @@
 from miqa.learning.nn_classifier import evaluate1
 
 
-def evaluate_data(images: List[Image], session: Session):
+@app.task
```
zachmullen, line 29: There's a preferred shortcut for this:

```python
from celery import shared_task
```

Then you can just decorate with `@shared_task`, and they'll be automatically picked up by the app. https://docs.celeryproject.org/en/stable/django/first-steps-with-django.html#using-the-shared-task-decorator

miqa/core/tasks.py
```diff
@@ -45,7 +53,14 @@ def evaluate_data(images: List[Image], session: Session):
...
 
 
-def import_data(user, session: Session):
+@app.task
+def import_data(user_id, session_id):
+    try:
+        user = User.objects.get(id=user_id)
+        session = Session.objects.get(id=session_id)
+    except (User.DoesNotExist, Session.DoesNotExist):
+        raise ValueError('Provide the id of the user and the id of the session.')
```
zachmullen, line 62: We can remove this try/except too.

miqa/core/tasks.py
```diff
@@ -126,10 +141,16 @@ def import_data(user, session: Session):
...
-    evaluate_data(images, session)
+    evaluate_data.delay([image.id for image in images], session.id)
+
 
+@app.task
+def export_data(session_id):
+    try:
+        session = Session.objects.get(id=session_id)
+    except Session.DoesNotExist:
+        raise ValueError('Provide the id of the session.')
```
zachmullen, line 152: Remove this one too

miqa/learning/nn_classifier.py
```diff
@@ -211,7 +211,7 @@ def get_model(file_path=None):
 
     if file_path is not None:
         model.load_state_dict(torch.load(file_path, map_location=device))
-        print(f'Loaded NN model from file "{file_path}"')
+        # print(f'Loaded NN model from file "{file_path}"')
```
zachmullen, line 214: Is the inclusion of this intentional? Let's either delete the line of code or add it back in rather than leaving it commented. (Same for the other two changes in this file.)

zachmullen (changes requested):

miqa/learning/nn_classifier.py
dzenanz, line 214: Can we make this output (and a few more below) only active in development mode instead of commenting them out?

dzenanz (commented):
This type of evaluation might be inefficient. [Using a loader](https://github.com/OpenImaging/miqa/blob/05501da2dbbc13d7fd84627273c0ef28ef5cb07d/miqa/learning/nn_classifier.py#L454-L457) should be more efficient, overlapping IO and inference. Even if we disable workers, not having to re-load all the libraries for each image should provide some computation time savings.

miqa/learning/nn_classifier.py
zachmullen: I'd suggest we use the logging system rather than `print`, and set these at `debug` level if that's something we want.

dzenanz:
Maybe we could make `evaluateMany` function which takes a list of filenames and returns an array of dictionaries? `evaluate1` returns a single dictionary.

miqa/learning/nn_classifier.py
dzenanz: Is that active by default in PyCharm? If not, how do I enable it?

zachmullen: I think the logger pattern is better than using `if` and `print`, since it handles this conditional concept internally.

annehaley: I don't see that we've implemented a logger anywhere. Should we set it up?

zachmullen: The nice thing is, it's already set up! At the top of the file, just do:

```python
import logging

logger = logging.getLogger(__name__)
```

Then you can replace all the print statements `logger.debug`.

annehaley: Uncommented and replaced all print statements.

annehaley:
I added an attempt at an `evaluateMany` function. before changing tasks.py to invoke it, does this look like a good implementation, @dzenanz? It is just a slightly modified version of `evaluate1`.

miqa/learning/nn_classifier.py
```diff
@@ -302,6 +301,47 @@ def evaluate1(model, image_path):
...
+                    'info': torch.FloatTensor([0] * (regression_count + len(artifacts))),
+                }
+            ],
+            transform=train_transforms,
+        )
+        evaluation_loader = DataLoader(
+            evaluation_ds, batch_size=1, pin_memory=torch.cuda.is_available()
+        )
+
+        tensor_output = evaluate_model(model, evaluation_loader, device, None, 0, 'evaluate1')
```
dzenanz, line 332:
```suggestion
        tensor_output = evaluate_model(model, evaluation_loader, device, None, 0, 'evaluateMany')
```

miqa/learning/nn_classifier.py
```diff
@@ -302,6 +301,47 @@ def evaluate1(model, image_path):
...
+            ScaleIntensityd(keys=['img']),
+            ToTensord(keys=['img']),
+        ]
+    )
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+    results = {}
+    for image_path in image_paths:
+        evaluation_ds = monai.data.Dataset(
+            data=[
```
dzenanz, line 320: You should construct `data` in a for loop, then do `evaluation_ds = monai.data.Dataset(` and the rest of this just once.

miqa/learning/nn_classifier.py
```diff
@@ -302,6 +301,47 @@ def evaluate1(model, image_path):
...
+                }
+            ],
+            transform=train_transforms,
+        )
+        evaluation_loader = DataLoader(
+            evaluation_ds, batch_size=1, pin_memory=torch.cuda.is_available()
+        )
+
+        tensor_output = evaluate_model(model, evaluation_loader, device, None, 0, 'evaluate1')
+        result = tensor_output.cpu().tolist()[0]
```
dzenanz, line 333: Once you have the results, they should be for the entire batch (all the files in `image_paths`).

miqa/learning/nn_classifier.py
```diff
@@ -302,6 +301,47 @@ def evaluate1(model, image_path):
...
+        labeled_results = {
```
dzenanz, line 334: This part should be done in a for loop too.

dzenanz (changes requested):

miqa/learning/nn_classifier.py
annehaley: cool, I didn't know we could do that. Is this second attempt what you meant?

miqa/learning/nn_classifier.py
```diff
@@ -314,32 +314,33 @@ def evaluateMany(model, image_paths):
...
-                    'img': image_path,
-                    'info': torch.FloatTensor([0] * (regression_count + len(artifacts))),
-                }
-            ],
-            transform=train_transforms,
-        )
-        evaluation_loader = DataLoader(
-            evaluation_ds, batch_size=1, pin_memory=torch.cuda.is_available()
-        )
+    evaluation_ds = monai.data.Dataset(
```
dzenanz, line 317: The previous version was closer to working. Let me write this function. I will push it to this fork.

annehaley: Ah okay, thanks. I was a bit confused about what you had in mind; I'm not familiar with the usage of these monai and torch classes.

dzenanz:
I tested `evaluate1` and `train_and_save_model` (which implicitly calls `evaluate_model`), but I expect the new function `evaluate_many` to work too.

miqa/learning/nn_classifier.py
```diff
@@ -17,6 +17,10 @@
 from torch.utils.tensorboard import SummaryWriter
 import wandb
 
+log_level = os.environ.get('LOGLEVEL', 'WARNING').upper()
+logging.basicConfig(level=log_level)
```
zachmullen, line 21: I think we should move the above two lines into the `__main__` block, for whoever is using this from the command line.

miqa/learning/nn_classifier.py
```diff
@@ -245,9 +249,9 @@ def evaluate_model(model, data_loader, device, writer, epoch, run_name):
                 print('', flush=True)
 
         if writer is not None:  # this is not a one-off case
-            print('\n' + run_name + '_confusion_matrix:')
-            print(confusion_matrix(y_true, y_pred))
-            print(classification_report(y_true, y_pred))
+            logger.info('\n' + run_name + '_confusion_matrix:')
```
zachmullen, line 252: Let's remove the `\n`, in the spirit of logging.

miqa/learning/nn_classifier.py
```diff
@@ -302,6 +305,45 @@ def evaluate1(model, image_path):
...
+    labeled_results = {}
+    for index, result in enumerate(results):
+        corresponding_image_path = image_paths[index]
+        single_labeled = {
+            'overall_quality': result[0],
+            'signal_noise_ratio': result[1],
+            'contrast_noise_ratio': result[2],
+        }
+        for index, artifact_name in enumerate(artifacts):
+            single_labeled[artifact_name] = result[index + 3]
```
zachmullen, line 342: Alternatively, you could do:

```suggestion
        for artifact_name, value in zip(artifacts, result[regression_count:]):
            single_labeled[artifact_name] = value
```

zachmullen (dismissed):

PR #155, opened 2021-09-21
Rename session to project

dchiquito:
This is, by necessity, a very large diff. Basically it's just a find/replace of `session`/`Session` with `project`/`Project`. The only things that weren't replace were the message asking the user if they want to continue their session in the timeout dialog, some comments in `settings.py` regarding OAuth sessions, and the historical migrations that set up the original `Session`.

The interesting bit to review is the migration, which I had to adjust manually in order to copy the data. In reality our risk is probably quite low, since A) production doesn't exist yet, and B) we wipe out sessions on import regularly anyway, but I did my best.


annehaley (approved):
This seems pretty straightforward, and everything appears to work when I run it. Just like you, I'm hoping that the evidence of Sessions in our migration files won't mess anything up in the future, but I doubt it will.

PR #159, opened 2021-09-22
Split the NN code into two files, training and inference

dzenanz:
I only tested the training and single image evaluation. It would be good if @annehaley tested the evaluation on import.


dzenanz:
I finally got my linting-fu fixed locally. This should be ready for review now.

annehaley (approved):
Everything works for the evaluations AFAICT

PR #167, opened 2021-09-29
Add new samples with brain images

annehaley:
This PR is intended to add new images to the samples directory so that we can load more complex experiment structures into MIQA and use real brain images (some of which are intentionally defected for NN evaluation purposes). These scans are a subset of the PredictHD set, reorganized to fit the import specifications set forth by MIQA. The file `samples/predictHD_tiny.csv` is the import csv to use for this set.


annehaley:
closing because we should use a subset of a different dataset

PR #168, opened 2021-09-29
Rename project scan

dchiquito:
Basically a big find/replace for "project" to "scan". Originally in the frontend scans were referred to as "sessions", as in a person having a session in an MRI. "session" was find/replaced to "project" in an earlier PR. I need to do this renaming now so I can add stuff for multi project support.


annehaley:
so we aren't renaming the model, just correcting the nomenclature used in the frontend?

annehaley (approved):
This makes sense to me and will clear some of the confusion when translating between frontend and backend.

PR #181, opened 2021-10-05
Update import method

annehaley:
right now, the import function treats scan_type as a unique identifier and won't allow two scans of the same type in an experiment.This is likely meant to be more like scan_id or scan_name. It also uses hard-coded column names and looks for the file location by asking for a root folder and then intuiting the location from there using the other info about the file. The import csv should just specify the whole file location in each line (relative to the samples root).

This PR involves major changes and should be looked at with scrutiny; the import function was rewritten using pandas and some models were changed to reflect the new assumptions we make about the data.

This PR also adds a file in `miqa/samples` which will allow us to easily convert from the old CSV format to the new CSV format by simply specifying some additional information that was not clear in the old format.


annehaley:
This is set as a draft for now, because the majority of tests will fail. The import code was rewritten, so the tests for that code will need to be updated.

miqa/core/models/scan.py
```diff
@@ -5,21 +5,22 @@
...
 
 class Scan(TimeStampedModel, models.Model):
     class Meta:
-        constraints = [
-            models.UniqueConstraint(
-                fields=['experiment', 'scan_id', 'scan_type'], name='scan_unique_constraint'
-            )
-        ]
-        ordering = ['scan_id']
+        ordering = ['name', 'scan_type']
```
dzenanz, line 21: Can different timesteps have different `scan_type`? If not, why order by it? If yes, I assume it is only due to convenience, despite us not expecting it?

dzenanz (commented):
Looks fine to me

miqa/core/models/scan.py
annehaley: This ordering is set on the `Scan` model, not the `Image` model. We do not expect Images/Frames to have different scan types, but different scans in the same experiment (possibly even with the same name) can have different scan types.

curtislisle:
Thanks for looking at this.  I didn’t like the rigdity of the previous import process, regarding names columns, the strange scan-id numbers, and the file name requirements.

miqa/core/tasks.py
```diff
@@ -56,90 +44,92 @@ def evaluate_data(image_ids, project_id):
...
+                }
+                all_scans.append(these_scans)
+                for scan_name, scan_object in these_scans.items():
+                    these_frames = {}
+                    for row in df[df['scan_id'] == scan_name].iterrows():
+                        raw_path = row[1]['file_location']
+                        if raw_path[0] != '/':
+                            # not an absolute file path; refer to project import csv location
+                            # TODO: add support for interpreting URIs not on host machine
+                            raw_path = str(Path(project.import_path).parent.parent) + '/' + raw_path
```
dchiquito, line 107: This `.parent.parent` seems quite arbitrary. Could we not use a Django setting that sets a global file root that any relative paths are relative to? In production there is a `MIQA_MOUNT_DIR` that specifies a directory to mount that presumably contains files for importing, and in dev there is `SAMPLES_DIR` that does basically the same thing. We could unify those settings or use the Django media storage or something.

miqa/core/tasks.py
```diff
@@ -56,90 +44,92 @@ def evaluate_data(image_ids, project_id):
...
+                            scan=scan_object,
+                            raw_path=raw_path,
+                        )
+                    all_frames.append(these_frames)
+
+        for model, list_all in [
+            (Project, all_projects),
+            (Experiment, all_experiments),
+            (Scan, all_scans),
+            (Image, all_frames),
```
dchiquito, line 119:
```
            (Experiment, new_experiments),
            (Scan, new_scans),
            (Image, new_frames),
```
Since these are apparently only used to create newly imported models, I think calling them `new_` rather than `all_` is a little more legible.

miqa/core/tasks.py
```diff
@@ -56,90 +44,92 @@ def evaluate_data(image_ids, project_id):
...
+        ]:
```
dchiquito, line 120: This loop thing is clever, but seems maybe overcomplicated. If you instead add items to `all_experiments` et. al. like this:
```
all_experiments.append([ experiment for _name, experiment in these_experiments.items() ])
```
You can just call `Experiment.objects.bulk_create(all_experiments)` directly (same for project, scan, Image). This also avoids the `if model == Image:` gotcha, you can just call `evaluate_data.delay` directly.

miqa/core/tasks.py
```diff
@@ -56,90 +44,92 @@ def evaluate_data(image_ids, project_id):
...
+            all_items = [item for subdict in list_all for item in subdict.values()]
+            model.objects.bulk_create(all_items)
+            if model == Image:
+                # TODO: modify this to support multi-projects, too
+                evaluate_data.delay([item.id for item in all_items], project.id)
+                # evaluate_data([item.id for item in all_items], project.id)
+
+    # TODO: re-write JSON parsing to be generalized like CSV parsing
+    # elif project.import_path.endswith('.json'):
+    #     pass
```
dchiquito, line 130: Are you planning to tackle JSON parsing in this PR?

samples/convert_miqa_import_csv_format.py
```diff
@@ -0,0 +1,65 @@
+import pandas
+from pathlib import Path
+import sys
+
+for file_name in sys.argv[1:]:
```
dchiquito, line 5: I've been quite impressed with [djclick](https://github.com/GaretJax/django-click) for writing scripts like this. It basically lets you register your own manage.py commands, which isn't immediately useful for this since you don't need the django settings or models or anything, but it might be a good central place to keep scripts.

See https://github.com/dandi/dandi-api/tree/master/dandiapi/api/management/commands for some examples

miqa/core/tasks.py
zachmullen: FYI I think you want `extend` rather than `append`

miqa/core/models/scan.py
annehaley: Ok, changing to just order by name.

miqa/core/tasks.py
```diff
@@ -56,90 +44,92 @@ def evaluate_data(image_ids, project_id):
...
+                    'experiment_id'
+                ].unique()
+            }
+            all_experiments.append(these_experiments)
+            for experiment_name, experiment_object in these_experiments.items():
+                these_scans = {
+                    scan_name: Scan(
+                        name=scan_name, scan_type=scan_type, experiment=experiment_object
+                    )
+                    for scan_name, scan_type in df[df['experiment_id'] == experiment_name]
```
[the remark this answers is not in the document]
annehaley: We can simply make the boolean filter a double: `... in df[df["experiment_id"] == experiment_name and df["project_id"] == project_name]`

samples/convert_miqa_import_csv_format.py
```diff
@@ -0,0 +1,65 @@
...
+    expected_columns = [
+        "xnat_experiment_id",
+        "nifti_folder",
+        "scan_id",
+        "scan_type",
+        "experiment_note",
+        "decision",
+        "scan_note",
+    ]
+    if len(df.columns) != len(expected_columns) or any(df.columns != expected_columns):
```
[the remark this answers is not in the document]
annehaley: Adding suggestion to batch.

miqa/core/tasks.py
annehaley: This is a pretty big change. Should we make an issue and put this on another PR?

annehaley:
This is not as much of a problem as you might think. We still definitely want the `scan_name` column to be the distinguishing field between scans of the same type in the same experiment. Before, we constrained against having that. So we just need to change a few things to add these types to the list. 

While I was at it, I changed the column titles to use `_name` instead of `_id`, since you made a good point about that above.

PR #202, opened 2021-10-26
Use NN with 5 convolutional blocks

dzenanz:
This reduces the variability between different validation folds, by noticeably improving the worst two cases. It also mildly reduces precision of one of the good folds.


dzenanz:
@annehaley Can you verify that this branch works as expected? If so, you could merge it.

Issue #208, opened 2021-10-28
Automatic population of reviewer's comment box with NN results


annehaley:
The control panel redesign concept included a feature wherein the results of an automatic evaluation from the server NN would automatically populate the reviewer's comment box so they could either agree or disagree with those results. However, this feature needs more planning and was therefore left out of the control panel redesign effort.

We need to consider the following: 

- What should the backend syntax for these comments on a ScanDecision be? How should we represent what features the user agrees are present? Should we include the percentage of confidence presented by the NN Evaluation? 
- How can we use this comment syntax to later feed back into the NN Model for continuous training?
- How should the UI design look for these tags, and how can we make it clear that a user can add additional commentary? Should the additional commentary go before or after the automatic evaluation representation?

curtislisle:
In our discussion on 12/2/21, we started focusing on an additional field for the ScanDecision model to support other binary tags attached to scans.  The NN results agreed-on by the user could be stored here as well as annotation tags users apply to scans.  Adding a comment here to address this with this issue or another related issue to be created.

PR #209, opened 2021-10-28
Control panel redesign (3/3)

annehaley:
The target bugs for this final PR of the control panel redesign are as follows:

- The initial for the decision made on a scan (displayed in the Experiment List as G, B, or O) should use the last decision made on that scan.
- Automatic locking and unlocking of an experiment when a reviewer opens it should work with no unexpected behavior, especially when two reviewers are interested in the same experiment.
- When an Experiment note is updated and successfully saved, the user may flip to the next scan or frame and will see the previous value again
- When a user is flipping through frames, the values for window width and window level are adjusted automatically as a guess based on the attributes of the frame. However, if the user manually sets either of these values, they should not be automatically adjusted for the rest of the scans.

Additionally, one final feature was planned for the control panel redesign effort, but it will need more consideration before implementation. The redesign concept included a feature wherein the results of an automatic evaluation from the server NN would automatically populate the reviewer's comment box so they could either agree or disagree with those results. This feature should be saved for future implementation, so it is mentioned in new issue #208.

### DIFF
--- a/client/src/components/ControlPanel.vue
+++ b/client/src/components/ControlPanel.vue
@@ -157,7 +157,8 @@ export default {
     async handleExperimentNoteSave() {
       if (this.newExperimentNote.length > 0) {
         try {
-          await djangoRest.setExperimentNote(
+          const { updateExperiment } = store.commit;
+          const newExpData = await djangoRest.setExperimentNote(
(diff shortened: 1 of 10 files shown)


dchiquito (approved):
looks good to me

PR #210, opened 2021-11-02
Crosshairs

### LINKED ISSUE #170: Crosshair to show the other slice locations
This is something came from Jim and he mentioned that all other viewers have it. 

@annehaley fyi

annehaley:
This PR adds colored crosshairs to the VtkViewer components.

I worked with Forrest Li to use VTK internals for proper visualization of the crosshairs. The last two commits on this PR are the results of live pair programming sessions with Forrest to ensure that rendering is done properly.

Resolves #170

### DIFF
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -23,7 +23,7 @@ export default {
     screenshotContainer: document.createElement('div'),
   }),
   computed: {
-    ...mapState(['proxyManager', 'loadingDataset', 'xSlice', 'ySlice', 'zSlice']),
+    ...mapState(['proxyManager', 'loadingDataset', 'xSlice', 'ySlice', 'zSlice', 'iIndexSlice', 'jIndexSlice', 'kIndexSlice']),
     ...mapGetters(['currentDataset', 'currentScan']),
     representation() {
       return (
@@ -67,7 +67,16 @@ export default {
     slice(value) {
       this.representation.setSlice(value);
       if (this.setCurrentVtkSlices) {
+        const ijkMapping = {
+          x: 'i',
+          y: 'j',
+          z: 'k',
+        };
         this.setCurrentVtkSlices({ axis: this.name, value: this.roundSlice(value) });
+        this.setCurrentVtkIndexSlices({
+          indexAxis: ijkMapping[this.name],
+          value: this.representation.getSliceIndex(),
+        });
       }
     },
     xSlice() {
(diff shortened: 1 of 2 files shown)


dchiquito (approved):
I am not well versed in the vtkjs stuff, but the vue stuff looks good to me.

client/src/components/VtkViewer.vue
```diff
@@ -85,12 +104,31 @@ export default {
   mounted() {
     this.initializeView();
     this.initializeSlice();
+    this.updateCrosshairs();
+    this.renderSubscription = this.view.getInteractor().onRenderEvent(() => {
+      this.updateCrosshairs();
+    });
+    this.resizeObserver = new window.ResizeObserver((entries) => {
+      if (entries.length === 1) {
```
zachmullen, line 112: What are the `entries` here?

client/src/components/VtkViewer.vue
```diff
@@ -241,7 +391,7 @@ export default {
   bottom: 0;
   left: 0;
   right: 0;
-  background: linear-gradient(#3a3a3a, #1d1d1d);
+  background: gray;
```
zachmullen, line 394: Is this change intentional or leftover from debugging?

zachmullen (approved):
LGTM, just a couple questions

client/src/components/VtkViewer.vue
annehaley: Haha I had to ask Forrest the same thing. Part of the `ResizeObserver` [API](https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API) explains that the observer can observe multiple elements that got resized with the same action. This line was his recommendation.

client/src/components/VtkViewer.vue
annehaley: good catch, I meant to change that to solid black so the image looks seamless against the backdrop. I think it looks cleaner than the linear gradient. What do you think?

client/src/components/VtkViewer.vue
zachmullen: To make sure the logic matches the intent... if we receive one of these events that has multiple entries, we intentionally want to ignore it?

annehaley: Yes, we tell the observer to only observe one element per instance of the `VtikViewer` component. We don't want to execute this block for any entries other than the specified target `this.$refs.viewer`.
We have three instances of the `VtkViewer` component; without this `if`, this block would execute 3^2 times instead of 3 times.

PR #213, opened 2021-11-04
Make NN outputs a percentage in 0.0 to 1.0 range

dzenanz:


annehaley (approved):
Looks good, we can add this into `ui-redesign-staging` then into `master` from there. Thanks for doing this so quickly.

annehaley:
Screenshot from 2021-11-04 11-38-41

Looking good!

PR #217, opened 2021-11-05
Rename Image as Frame

### LINKED ISSUE #176: Rename Image model to Frame

annehaley:
Resolves #176


client/src/store/index.ts
```diff
@@ -387,13 +388,13 @@ const {
...
-      state.currentDatasetId = imageId;
+    setCurrentFrameId(state, frameId) {
+      state.currentDatasetId = frameId;
     },
-    setImage(state, { imageId, image }) {
+    setFrame(state, { frameId, frame }) {
       // Replace with a new object to trigger a Vuex update
       state.datasets = { ...state.datasets };
-      state.datasets[imageId] = image;
+      state.datasets[frameId] = frame;
```
zachmullen, line 397: In the spirit of renaming things, should this object be called `frames` rather than `datasets`, since that's what it is storing?

miqa/core/migrations/0018_image_to_frame.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0017_remove_site'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
zachmullen, line 16: I think you may have to hand-edit this migration to make it use a `migrations.RenameModel`. As written, this will delete all existing images in the system.

zachmullen (changes requested):

miqa/core/migrations/0018_image_to_frame.py
zachmullen: See https://stackoverflow.com/a/26241640/2550451

annehaley: Good idea. I've been wondering for a while, though, if there's ever a good time to squash all migrations to fewer files, since a lot of our current migration files go back and forth. Is that a good idea? Or do we want to keep all of these previous migration generations?

client/src/store/index.ts
annehaley: Good Idea; adding to commit.

miqa/core/migrations/0018_image_to_frame.py
annehaley: I hadn't thought of that. Adding change to next commit.

dchiquito (approved):

PR #218, opened 2021-11-08
Use reverse relation to access the Evaluation for a Frame

### LINKED ISSUE #199: Add related_name to Evaluations.image
This looks like it could be included on the `Image` model using [related_name](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey.related_name)
```suggestion
            evaluation_object = image_object.evaluations.get()
```

_Originally posted by @dchiquito in https://github.com/OpenImaging/miqa/pull/196#r731968289_

https://github.com/OpenImaging/miqa/pull/196/files#diff-4e38f01091ea4cb45dd3c10ad6650a542c31c960ca4ad4505a095752449e71c6R44 should be updated to use the `related_name` instead of querying `Evaluations.objects` directly.

Also, add `unique=True` to Evaluations.image. Every image can have at most one Evaluation, so it would be good to enforce that at the schema level.

annehaley:
Resolves #199


miqa/core/models/evaluation.py
```diff
@@ -5,7 +5,12 @@
 
 class Evaluation(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
-    frame = models.ForeignKey('Frame', null=False, on_delete=models.PROTECT)
+    frame = models.OneToOneField(
+        'Frame',
+        null=False,
+        on_delete=models.PROTECT,
```
zachmullen, line 11: Why `PROTECT` and not `CASCADE`?

annehaley: perhaps I have their behavior reversed, but I thought that the desired outcome was that deleting an Evaluation would not delete the target Frame, but deleting a Frame would delete all related Evaluations. Which is best to use for that?

zachmullen: When you set `on_delete` on Django ForeignKeys, it roughly mirrors the semantics of `ON DELETE` on the RDBMS foreign keys, meaning it refers to what happens when the _referenced_ object is deleted. So in this case, if you want deleting a Frame to delete all associated Evaluations, you want `models.CASCADE`. In the other direction, nothing will happen regardless of this value.

* `PROTECT` would mean that an attempt to delete the associated Frame would raise an exception and fail.
* `SET_NULL` would mean that deleting a Frame would preserve all the evaluations that were linked to it, but set their `frame` column to null.

annehaley: I see, thanks for the clarification. I think I always get them switched because I think of what happens when the _defined_ object is deleted. I'll make that change.

miqa/core/models/evaluation.py
```diff
@@ -5,7 +5,12 @@
 
 class Evaluation(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
-    frame = models.ForeignKey('Frame', null=False, on_delete=models.PROTECT)
+    frame = models.OneToOneField(
+        'Frame',
+        null=False,
```
zachmullen, line 10:
```suggestion
```

This is the default value, and should kind of be our mental default as well; nullability is the exception to the rule, so I'd prefer to only specify the argument when it's `True`.

annehaley: Good point!

dchiquito (approved):

PR #227, opened 2021-11-11
Permissions api

### LINKED ISSUE #71: Support multi-users in MIQA with permissions
Currently we have one user, we need to support multi-user mode where each user will have their own session and may import their own CSV 

* A user has multiple sessions to review
* A session may have only one user with write access and others may have read access (have a lock on the session)
* There is no need for concurrent editing
* Each user has equal rights to write access on any imported session as long as they have write permission
* For now, any user can unlock the session but will be prompted 
* We should be able to see who owns the lock

annehaley:
This PR is scoped only for the API implementation of user roles on projects. UI changes will have a separate PR.

Resolves #71 . This PR includes the following: 

- Inclusion of specialized permission groups on the Project object: collaborators, tier 1 reviewers, and tier 2 reviewers (dynamic tiering not implemented; that can be a future upgrade)
- Implementing functionality in the `PUT /projects/{id}/settings` endpoint to change the users in each permission group, using a function on the `Project` model
- Adding a decorator for our API views to check for a given permission level and invoking that decorator to enforce permission rules according to what is illustrated in [this Google Sheet](https://docs.google.com/spreadsheets/d/1PFOMpDkYJLoDIcLP9Ilts4jX5_3sd3bbJMkT7M4Uzik/edit?usp=sharing)
- Rewriting the pytest suite to include user permissions as another variable dimension by parametrizing the `user_api_client` fixture


annehaley:
Right now, this will not pass the pytests because they are broken by these changes. This PR will soon include commits to change the pytests.

miqa/core/rest/permissions.py
```diff
@@ -1,12 +1,58 @@
...
 from rest_framework.request import Request
+from rest_framework.response import Response
 from rest_framework.views import View
 
-from miqa.core.models import Experiment
+from miqa.core.models import Experiment, Project
+
+
+def project_permission_required(
+    read_access=True, review_access=False, superuser_access=False, **kwargs
```
dchiquito, line 17: These arguments have 8 different permutations, but only three are used (or even meaningful). What do you think about having a different decorator for each permission level, like `@project_read_access_required` and `@project_review_access_required`?
Properly abstracting the different decorators would involve a function that takes a function and generates a decorator, which is perhaps just too silly, but it would make the API a little cleaner.

Incidentally, I feel like there is an existing decorator

miqa/settings.py
```diff
@@ -40,8 +40,17 @@ def before_binding(configuration: ComposedConfiguration) -> None:
         # Install additional apps
         configuration.INSTALLED_APPS += [
             's3_file_field',
+            'guardian',
         ]
 
+        # guardian's authentication backend
+        configuration.AUTHENTICATION_BACKENDS += [
+            'guardian.backends.ObjectPermissionBackend',
```
dchiquito, line 48: What does this do?

miqa/core/models/project.py
```diff
@@ -61,3 +62,30 @@ def clean(self):
...
+        for username in user_list:
+            new_permitted_user = User.objects.get(username=username)
+            if new_permitted_user not in old_list:
+                assign_perm(group_name, new_permitted_user, self)
+
+    class Meta:
+        permissions = (
+            ('collaborators', 'Collaborators'),
+            ('tier_1_reviewers', 'Tier 1 Reviewers'),
+            ('tier_2_reviewers', 'Tier 2 Reviewers'),
```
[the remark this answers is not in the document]
annehaley: Adding suggestion to batch.

miqa/core/rest/permissions.py
```diff
@@ -1,12 +1,58 @@
...
+                'FrameViewSet': 'experiments__scans__frames__pk',
+            }
+            lookup_dict = {lookup_mapping[viewset.__class__.__name__]: kwargs['pk']}
+            project = get_object_or_404(Project, **lookup_dict)
+
+            user = viewset.request.user
+            user_perms_on_project = get_perms(user, project)
+            has_review_perm = all(
+                [
+                    perm not in user_perms_on_project
```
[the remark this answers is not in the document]
annehaley: No, I was mistaken. I wrote it one way and realized the boolean logic was too much, so I tried to demorgan's it and simplify, but I failed to catch all the "not"s. Fixing in batch.

miqa/core/rest/permissions.py
```diff
@@ -1,12 +1,58 @@
...
+    read_access=True, review_access=False, superuser_access=False, **kwargs
+):
+    def decorator(view_func):
+        def _wrapped_view(viewset, *args, **kwargs):
+            lookup_mapping = {
+                'ProjectViewSet': 'pk',
+                'ExperimentViewSet': 'experiments__pk',
+                'ScanViewSet': 'experiments__scans__pk',
+                'ScanDecisionViewSet': 'experiments__scans__decisions__pk',
+                'FrameViewSet': 'experiments__scans__frames__pk',
```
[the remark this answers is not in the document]
annehaley: That sounds good, adding to batch.

miqa/core/rest/permissions.py
```diff
@@ -1,12 +1,58 @@
...
+            }
+            lookup_dict = {lookup_mapping[viewset.__class__.__name__]: kwargs['pk']}
+            project = get_object_or_404(Project, **lookup_dict)
+
+            user = viewset.request.user
+            user_perms_on_project = get_perms(user, project)
+            has_review_perm = all(
+                [
```
zachmullen, line 34: You can remove the square braces; `all` is one of the special built-ins that can accept a generator expression syntax.

miqa/core/rest/permissions.py
```diff
@@ -1,12 +1,58 @@
...
+                    perm not in user_perms_on_project
+                    for perm in Project.get_review_permission_groups()
+                ]
+            )
+            has_read_perm = all(
+                [perm not in user_perms_on_project for perm in Project.get_read_permission_groups()]
```
zachmullen, line 40:
```suggestion
                perm not in user_perms_on_project for perm in Project.get_read_permission_groups()
```

Same here

zachmullen (dismissed):

miqa/core/rest/permissions.py
annehaley: We can just take off `read_access` as an arg, since we assume it to be true and would never use this decorator with it as False. That brings our 8 permutations down to 4.

miqa/settings.py
annehaley: Part of the configuration for guardian, I was just following [the guide in the docs](https://django-guardian.readthedocs.io/en/stable/configuration.html#:~:text=AUTHENTICATION_BACKENDS%20%3D%20(%0A%20%20%20%20%27django.contrib.auth.backends.ModelBackend%27%2C%20%23%20this%20is%20default%0A%20%20%20%20%27guardian.backends.ObjectPermissionBackend%27%2C%0A))

miqa/core/rest/permissions.py
annehaley: Oh yeah, you told me that a few PRs ago and I keep forgetting!

zachmullen (approved):
I'm not much of an expert on django guardian, so I can't give a terribly meaningful approval. However, the code looks generally good.

miqa/core/rest/scan_decision.py
```diff
@@ -23,23 +24,30 @@ class Meta:
...
+        )
+        return ScanDecision.objects.filter(scan__experiment__project__in=projects)
+
+    # cannot use project_permission_required decorator because no pk is provided
+    def create(self, request, **kwargs):
         request_data = request.data
-        request_data['scan'] = Scan.objects.get(id=request.data['scan'])
+        scan = Scan.objects.get(id=request.data['scan'])
+        if get_perms(request.user, scan.experiment.project) == []:
+            return Response(status=status.HTTP_401_UNAUTHORIZED)
```
dchiquito, line 47: This looks like it implicitly checks if the user has reviewer access or higher, but that's not really clear from the code at all, and seems rather fragile. It might be worth breaking the decorator into two parts: a helper `has_project_permission(user, project, review_access=False, superuser_access=False)`, and the decorator which calls it. Then you can call `has_project_permission` here or anywhere else you might want to do an explicit permission check.

miqa/core/tests/test_rest.py
```diff
@@ -1,92 +1,143 @@
...
+    resp = user_api_client.put(
         f'/api/v1/projects/{project.id}/settings',
-        data={'importPath': '/new/fake/path', 'exportPath': '/new/fake/path'},
+        data={
+            'importPath': '/new/fake/path',
+            'exportPath': '/new/fake/path',
+            'permissions': {
+                'collaborator': [],
+                'tier_1_reviewer': [user.username],
+                'tier_2_reviewer': [],
```
dchiquito, line 48: These seem like a good candidate for fixturization. Probably something like
```
@pytest.mark.fixture(
    'role',
    ['collaborator', 'tier_1_reviewer', 'tier_2_reviewer']
)
```
```suggestion
                'collaborator': [user.username] if role == 'collaborator' else [],
                'tier_1_reviewer': [user.username] if role == 'tier_1_reviewer' else [],
                'tier_2_reviewer': [user.username] if role == 'tier_2_reviewer' else [],
```

miqa/core/tests/test_rest.py
```diff
@@ -1,92 +1,143 @@
...
+        assert resp.status_code == 200
+        assert user_api_client.get(f'/api/v1/projects/{project.id}/settings').data == {
+            'importPath': '/new/fake/path',
+            'exportPath': '/new/fake/path',
+            'permissions': {
+                'collaborator': [],
+                'tier_1_reviewer': [user.username],
+                'tier_2_reviewer': [],
+            },
+        }
```
dchiquito, line 64: Can you add an assertion that `get_perms(...) == ['tier_1_reviewer']`? Good to make sure the intended side affect ocurred.

miqa/core/tests/test_rest.py
```diff
@@ -96,105 +147,154 @@ def test_experiment_lock_acquire_requires_auth(api_client, experiment):
...
         '/api/v1/scan-decisions',
         data={
             'scan': scan.id,
             'decision': 'Good',
             'note': 'hello',
         },
     )
-    assert resp.status_code == 403
-    assert resp.data['detail'] == 'You must lock the experiment before performing this action.'
+    if get_perms(user, scan.experiment.project) == []:
```
dchiquito, line 258: These checks would all benefit from a `has_project_permission` helper

miqa/core/rest/scan_decision.py
dchiquito: Also, breaking it up allows you to write a unit test for the `has_project_permission` function in isolation, which is generally a good idea when you have a big chunk of code like that. As it is I'm not sure how you would go about writing a test for `@project_permission_required` in isolation, although we do implicitly test it by testing the views that use it.

dchiquito (changes requested):
The logic looks good to me, I just had one ask for the `@project_permission_required` decorator and some minor quibbles for the tests.

miqa/core/rest/scan_decision.py
annehaley: Good point. I wan't able to turn it into a decorator because the pk of a db object is not passed through the parameters of this PUT request. However, I made functions shared by the decorator itself and the endpoint function. These functions have a more concrete way of testing for the proper permissions.

miqa/core/tests/test_rest.py
annehaley: Sorry, I don't understand where the first snippet goes to define role for the test function. Where should I put that?

miqa/core/tests/test_rest.py
annehaley: Sure, should I do an `if` block to assert the change made based on the `role` obtained in the previous change?

miqa/core/tests/test_rest.py
```diff
@@ -1,92 +1,143 @@
...
-def test_experiments_list(authenticated_api_client, experiment):
-    resp = authenticated_api_client.get('/api/v1/experiments')
+def test_experiments_list(user_api_client, experiment, user):
+    resp = user_api_client(project=experiment.project).get('/api/v1/experiments')
     assert resp.status_code == 200
-    assert resp.data['count'] == 1
+    if get_perms(user, experiment.project) == []:
+        assert resp.data['count'] == 0
+    else:
+        assert resp.data['count'] == 1
```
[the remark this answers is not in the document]
annehaley: Good point. I wrote a more complete assertion with hard-coded expected values.

miqa/core/tests/test_rest.py
annehaley: Agreed. I used the same helper functions in `permissions.py` that I implemented for your first suggestion.

miqa/core/tests/test_rest.py
annehaley: I was able to use a simpler `if` from the information we already had in the function, like this:

```
my_perms = get_perms(user, project)
expected_perms = {
        'collaborator': [user.username] if 'collaborator' in my_perms else [],
        'tier_1_reviewer': [user.username] if 'tier_1_reviewer' in my_perms else [],
        'tier_2_reviewer': [user.username] if 'tier_2_reviewer' in my_perms else [],
    }
```

miqa/core/tests/conftest.py
```diff
@@ -20,17 +23,21 @@ def api_client() -> APIClient:
...
+@pytest.fixture(params=[None, 'superuser'] + Project.get_read_permission_groups())
+def user_api_client(request, api_client, user, project) -> APIClient:
+    def _method(**kwargs):
+        api_client.force_authenticate(user=user)
+        if request.param:
+            if request.param == 'superuser':
+                user.is_superuser = True
+                user.save()
+            elif 'project' in kwargs:
+                assign_perm(request.param, user, kwargs['project'])
```
dchiquito, line 35: That's cool :+1:

dchiquito (approved):

PR #232, opened 2021-11-22
Deployment fixes

annehaley:
Two small fixes noted while helping @curtislisle set up his local environment

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -8,13 +8,13 @@ services:
     # Log printing via Rich is enhanced by a TTY
     tty: true
     env_file: ./dev/.env.docker-compose
-    deploy:
-      resources:
-        reservations:
-          devices:
-            - driver: nvidia
-              count: 1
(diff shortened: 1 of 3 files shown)


docker-compose.override.yml
```diff
@@ -8,13 +8,13 @@ services:
     # Log printing via Rich is enhanced by a TTY
     tty: true
     env_file: ./dev/.env.docker-compose
-    deploy:
-      resources:
-        reservations:
-          devices:
-            - driver: nvidia
-              count: 1
-              capabilities: [gpu]
```
dchiquito, line 17: Can you leave a note that this needs to be uncommented when we implement active learning (or whatever the technical term is)?

dchiquito (approved):

docker-compose.override.yml
annehaley: Sure. I believe active learning is the technical term.

PR #234, opened 2021-11-22
Fix console errors

dchiquito:
There were several race conditions where objects loaded asynchronously
would sometimes be null, causing console errors.


annehaley (approved):
All good catches. These should help prevent those errors popping up in the console.

PR #236, opened 2021-11-23
Fix example data

### LINKED ISSUE #235: Dataset view orientation inconsistent
To reproduce:
* Go to the projects list, select a project containing the sample dataset
* Click on Scan
* The viewers will be black, but crosshairs will be present.
* Click up/down to go to a different experiment
* The viewers will be black, and the crosshairs will be misplaced.
* Click down/up to return to the first experiment
* The viewers will load the image correctly, but one crosshair axis will be absent in each axial view.

annehaley:
Resolves #235.
Additional changes here fix the bug that the import and export paths did not appear in their tooltips as they should, as well as refactoring the cleanup on the `VtkViewer` component.

### DIFF
--- a/client/src/components/EvaluationResults.vue
+++ b/client/src/components/EvaluationResults.vue
@@ -27,7 +27,7 @@ export default {
         'green lighten-1',
         'green darken-2',
         'green darken-4',
-      ];
+        'black '];
       const thisColor = colors[Math.floor(Math.abs(value) * (colors.length - 1)) % colors.length];
       if (text) {
         return `font-weight-bold ${thisColor.replace(' ', '--text text--')}`;
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -1,9 +1,11 @@
 <script>
-import Vue from 'vue';
-import { mapState, mapGetters, mapMutations } from 'vuex';
 import { vec3 } from 'gl-matrix';
 
+import Vue from 'vue';
+import { mapState, mapGetters, mapMutations } from 'vuex';
+
 import { cleanFrameName } from '@/utils/helper';
+import CrosshairSet from '../utils/crosshairs';
 import fill2DView from '../utils/fill2DView';
 import { getView } from '../vtk/viewManager';
 import { VIEW_ORIENTATIONS } from '../vtk/constants';
@@ -25,7 +27,7 @@ export default {
     screenshotContainer: document.createElement('div'),
   }),
   computed: {
-    ...mapState(['proxyManager', 'loadingFrame', 'showCrosshairs', 'xSlice', 'ySlice', 'zSlice', 'iIndexSlice', 'jIndexSlice', 'kIndexSlice']),
+    ...mapState(['proxyManager', 'loadingFrame', 'showCrosshairs', 'iIndexSlice', 'jIndexSlice', 'kIndexSlice']),
     ...mapGetters(['currentFrame', 'currentScan']),
     representation() {
       return (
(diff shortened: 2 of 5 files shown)


dzenanz (commented):
You probably can't reuse this code from ITK:
https://github.com/InsightSoftwareConsortium/ITK/blob/8801247049712d50528e4651a0d000892b7eb22a/Modules/Core/Common/src/itkSpatialOrientationAdapter.cxx#L32-L111
but you might reuse the logic.
Since there is the option of getting the direction matrix with ones from the orientation code, the flow could be like this:
1. get orientation code
2. get orientation matrix with 1s
3. replace the current orientation by the one with 1s
4. show the image using the original logic (before trying to orthogonalize it using the current, buggy approach)

dzenanz:
@annehaley I just added some test data with several common and a few uncommon orientations
https://drive.google.com/drive/u/1/folders/1AmpmA_L_TM3r0Zvrtimsi0msl5ibfycI
I put all files in both `.nrrd` and `.nii.gz` formats.

dzenanz:
I am not sure whether this is the correct link to what Slicer does about it, but it possibly is:
https://github.com/Slicer/Slicer/blob/88dd2151af6272684e32d0c8b98a881f1b28b689/Libs/MRML/Core/vtkMRMLSliceNode.cxx#L2207-L2302

client/src/components/EvaluationResults.vue
```diff
@@ -27,7 +27,7 @@ export default {
         'green lighten-1',
         'green darken-2',
         'green darken-4',
-      ];
+        'black '];
```
dzenanz, line 30: This seems like a spurious change.

annehaley:
Some things worth noting on these changes: 

1. Oblique images are still not supported; Forrest explained that these would require image resampling on the frontend.
2. There are still some edge cases for the crosshairs now, but that can be resolved in another PR.

client/src/components/VtkViewer.vue
```diff
@@ -40,8 +40,14 @@ export default {
     name() {
       return this.view.getName();
     },
+    closestIJKAxis() {
+      return this.representation.getMapper().getClosestIJKAxis().ijkMode;
+    },
+    closestIJKFlip() {
```
dzenanz, line 46: I don't see this used anywhere.

client/src/components/VtkViewer.vue
```diff
@@ -40,8 +40,14 @@ export default {
     name() {
       return this.view.getName();
     },
+    closestIJKAxis() {
```
dzenanz, line 43: not used?

dzenanz (commented):
Is this working now?

dzenanz:
> Oblique images are still not supported; Forrest explained that these would require image resampling on the frontend.

I don't think we need resampling. Oblique images are quite rare, and it is fine if we show them in the closest orthogonal orientation.

client/src/components/EvaluationResults.vue
annehaley: Oh yes, I added that because I noticed on this set that the `EvaluationResults` component was rendering wrong for soem of our example data (which represented an edge case for the Evaluation Results). This change does actually affect the appearance of that component in certain cases.

client/src/components/VtkViewer.vue
annehaley: Good catch. Removing.

dzenanz (approved):
Preferably to be squashed upon merge.

dzenanz:
@curtislisle can you try this branch? It should be handling all orientations (except some oblique cases).

curtislisle:
I tried to pull and test this PR, but the node container quits when I do a compose up with an OpenSSL error:

```
  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
  library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'
```
I believe from stackoverflow, this is because of node 17...(my error is exactly what is mentioned here): [stackoverflow error link](https://stackoverflow.com/questions/69394632/webpack-build-failing-with-err-ossl-evp-unsupported)

The npm container says it is running Node.js v17.3.0,  do we need to specify the version of node in `docker-compose.override.yml` instead of `node:latest`?

zachmullen:
I had to fix this locally as well, I changed it from `node:latest` to `node:16` and it worked.

curtislisle:
@zachmullen , thanks.  I just found changing to node:lts worked as well, but I will switch to node:16 instead to pin the version.

dzenanz (approved):
We can handle additional cases later. If this is clearly an improvement and does not regress anything, we should merge it now.

dzenanz (approved):
Code looks good on a glance. I assume it works 😄

Issue #23, opened 2021-06-09
PredictHD: predict overall QA from other columns


dzenanz:
We should check correlation between overall QA score and presence of 10 artifacts in PredictHD QC data.

dzenanz:
Result of `df.corr()`:

  | normal_variants | lesions | snr | cnr | full_brain_coverage | misalignment | inhomogeneity | flow_artifact | truncation_artifact | overall_qa_assessment
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
normal_variants | 1.000 | 0.334 | -0.150 | -0.139 | -0.052 | 0.031 | 0.091 | 0.359 | 0.273 | -0.133
lesions | 0.334 | 1.000 | -0.083 | -0.082 | -0.053 | 0.053 | 0.074 | 0.081 | 0.107 | -0.091
snr | -0.150 | -0.083 | 1.000 | 0.826 | 0.413 | -0.242 | -0.299 | -0.149 | -0.314 | 0.721
cnr | -0.139 | -0.082 | 0.826 | 1.000 | 0.414 | -0.219 | -0.297 | -0.153 | -0.297 | 0.715
full_brain_coverage | -0.052 | -0.053 | 0.413 | 0.414 | 1.000 | -0.103 | -0.124 | -0.003 | -0.035 | 0.723
misalignment | 0.031 | 0.053 | -0.242 | -0.219 | -0.103 | 1.000 | 0.285 | -0.053 | 0.137 | -0.248
inhomogeneity | 0.091 | 0.074 | -0.299 | -0.297 | -0.124 | 0.285 | 1.000 | 0.015 | 0.299 | -0.331
flow_artifact | 0.359 | 0.081 | -0.149 | -0.153 | -0.003 | -0.053 | 0.015 | 1.000 | 0.373 | -0.140
truncation_artifact | 0.273 | 0.107 | -0.314 | -0.297 | -0.035 | 0.137 | 0.299 | 0.373 | 1.000 | -0.286
overall_qa_assessment | -0.133 | -0.091 | 0.721 | 0.715 | 0.723 | -0.248 | -0.331 | -0.140 | -0.286 | 1.000

Unfortunately, this drops columns with some missing values. I will look into including those columns too.

dzenanz:
Here is correlation for all columns:

[...]

PR #237, opened 2021-11-29
Remove SNR and CNR from training and inference

dzenanz:
See analysis in #23.


dzenanz:
@annehaley as this removes 2 columns out of 13, some adjustments in UI will be needed. Feel free to push them to this branch (`master` in my fork).

annehaley:
> @annehaley as this removes 2 columns out of 13, some adjustments in UI will be needed. Feel free to push them to this branch (`master` in my fork).

Which UI changes do you mean? The current display component for Evaluation results does not include SNR or CNR.

dzenanz:
Ah, you might be right. Can you check that it still works?

annehaley:
Absolutely, I will be able to review it in a bit.

dzenanz:
Judging by [regression_count = 1](https://github.com/dzenanz/miqa/blob/95fe0a56c4ef995d01f0cd53810f9fe99cc3a9fb/miqa/learning/nn_inference.py#L13) and [classes=regression_count + len(artifacts)](https://github.com/dzenanz/miqa/blob/95fe0a56c4ef995d01f0cd53810f9fe99cc3a9fb/miqa/learning/nn_inference.py#L79), the current model should have final layer size of 11, and that should be in the weight files too. Can you double check for gross errors (e.g. failures of git checkout)? If that does not turn up anything, we could have a meeting to debug it.

annehaley:
Ah, you were right. It always seems to be a git-lfs problem. Once I was able to properly check out the changed model files, everything worked smoothly.

annehaley (approved):

PR #239, opened 2021-11-30
switch to native promises

### LINKED ISSUE #138: Remove Bluebird promises from web client

annehaley:
Resolves #138


annehaley:
After discussion with @dchiquito about these changes, we realized that bluebird was initially included to account for request cancellation when the user navigates away from a dataset they just requested. This cannot be accomplished with native promises. In order to determine if request cancellation makes a non-negligible difference in load times, I monitored the timing of  25 download requests for the 25 frames in IXI_small, comparing `remove-bluebird` and `master`. To do this, I pressed the down arrow key about once every half second. The results are shown below (respectively, where `remove-bluebird` is shown first). Three attempts for each looked fairly similar to these.

image
image

Both have a greater delay before the 22nd Frame load because the frontend takes a moment to switch between Experiments. It appears that `remove-bluebird` has more red, meaning more requests spend some time in a blocked state. However, the overall timing is about the same. It may be worthwhile to try this experiment in a more constrained manner and do some more in-depth analysis with larger files to load, but at the moment, it appears that bluebird makes no significant difference. 

@dchiquito do you think we should move forward on removing it from our frontend?

dchiquito:
Nice research. It looks like the cancellations aren't reducing network load at all. Even if there's some negligible JS performance improvement from cancelling the promise early, it looks like it doesn't impact the "feel" significantly.
I think it's worth running by the team at the next meeting, but to me the data seems to make a compelling argument to get rid of Bluebird.

dchiquito (approved):
Approved, pending team consensus

dzenanz (approved):
Having less dependencies is better. If the performance difference is negligible, I am in favor.

curtislisle:
Nice analysis.  Not having reviewed the code in question, is the observed timing likely to be relatively independent of dataset size?  In that case, native promises wouldn't be much slower even with a larger dataset.  If so, I approve too.

PR #241, opened 2021-12-01
Global import export

### LINKED ISSUE #224: Global CSV import and export
we should move to a global CSV import, potentially with an option for incremental update (but that is not priority). Even without incremental update of experiments, we will need projects to “stick around” on import - we don’t want imports to clear all the permissions/roles that have been set up

dchiquito:
Fixes #224 

I opted to add a toggle switch to the project settings page that enables/disables global import for that particular project.
When enabled, global imports will respect the project names specified in the import file, which can result in overwriting data in or exporting data from other projects.
When disabled, the project names in the import file are ignored completely, and the project the settings are defined for is used instead.

Global import/export is a stopgap to support the collaborators current use case, so we should remove it and redesign the import file to exclude project names as soon as we revamp the import process (probably with an XNAT integration).


client/src/components/ProjectSettings.vue
```diff
@@ -120,6 +124,25 @@ export default defineComponent({
...
+    <v-layout>
+      <v-tooltip top>
+        <template v-slot:activator="{ on, attrs }">
+          <div
+            v-bind="attrs"
+            v-on="on"
+          >
+            <v-switch
+              v-model="globalImportExport"
+              @click="changed = true"
```
annehaley, line 136: I tend to think that on a form, toggles should perform automatic saves. Unlike text fields where you can't tell when the user is finished, a toggle is a one-and-done action like a button. Perhaps the change function should call the save function instead of making the Save button enabled.

annehaley (changes requested):
My test of these changes worked well. I just have UI thoughts.

client/src/components/ProjectSettings.vue
dchiquito: I think that would be less intuitive, considering the proximity of the save button which enables immediately after clicking the toggle. Having text fields that require a manual save and a toggle which doesn't seems inconsistent to me.

annehaley: Okay, fair enough. I just thought it was worth mentioning.

annehaley (approved):

PR #244, opened 2021-12-03
Artifact selection

annehaley:
A first pass at implementation of the artifact selection feature. 

I made a concept document and mockups that I discussed with @curtislisle before we made a first pass at this implementation in a pair programming session. The following ideas are implemented here:

1. Model Representation
>  The ScanDecision model should include a new JSON field, `user_identified_artifacts`, which must have keys equivalent to the artifacts list defined in `learning/nn_inference.py` (or the negative versions of these artifacts). The values can be any of `[-1, 0, 1]`, where -1 is the default. A value of -1 can mean unreviewed, while 0 means the user determined the artifact is not present and 1 means the user determined the artifact is present. The values in this JSON can be used to feed back to the NN for dynamic learning.

2. UI Widget
> The User can select any number of artifacts as flags. The artifact options are phrased as negative, so a perfect scan requires no extra clicks. The widget is automatically populated with the artifacts that the NN Evaluation identifies with a certainty past a threshold percentage that we can initially set to 75%. The user can remove these chips or add more if they disagree with the assessment of the NN. The chips that are selected will be saved with their values as 1. The chips that are absent will be saved with their values as 0.

artifact_selection_mockup_v1

These ideas are implemented with interchangeable functionality in mind, so that Curt can repurpose this feature for image tagging. One additional implementation detail: I added a new API endpoint for getting configuration details from the backend to the frontend so the frontend doesn't need to make hard-coded assumptions about configurable features.


jeffbaumes:
This looks very good, the only thing I have a concern about is:

> The widget is automatically populated with the artifacts that the NN Evaluation identifies with a certainty past a threshold percentage that we can initially set to 75%.

I don't like that the field that should be user-validated annotations by default will have some non-user-validated stuff in it.

What I'd rather is that the 75% confidence things show up as chips to check for before anything is saved, maybe like "Ghosting Motion ? (x)" that shows in a different color. Clicking the (x) removes it (and signals to save it as a verified "0" value), clicking on Ghosting Motion accepts it and the "?" goes away (and signals to save it as a verified "1" value). In that way, the "?" stuff never goes in the DB as a user-verified artifact. Would something like that be doable?

annehaley:
Yes! That's a good idea. I'll work on that.

annehaley:
In that case, should the API view receive two lists of `confirmed_present` and `confirmed_absent`? The former would assign 1s and the latter would assign 0s, leaving the rest as -1s?
If so, how do we determine which ones the user really thinks are absent? If an artifact is not auto populated, they do not need to clear it and would not interact with it.

curtislisle:
One approach could be to have a region of “suggested annotations”, which is where the auto-population occurs.  Then the user can select any of them and selected ones will move to show up in a separate “present annotations” area instead.  The v-select with all attributes for the user to pick would be in the “present annotations” section, as well.  This approach would require a separate “suggested annotations" UI space, but might be easier to understand quickly than changing the button color/icon.  I am OK with either approach here. 

On the automatic 0, Jeff has a good point in the “questionable” case, but I am concerned the logic could get complicated for users if we aren’t careful. 

Another idea is that we could draw chips for all features, with only the positive (>75%) ones filled in, then it is obvious what the algorithm thinks.  The user could select the body of any filled button to “accept it” and move it to the “validated present features", click the “X” on any button to deny any measurement being recorded (leave it -1) or leave the unfilled buttons where they are (to confirm absence of this feature).  This is a solution without much complicated logic, but I think it would make for a crowded UI panel.  

I am not really in love with my idea above, but it tries to avoid having logic that might be confusing.

jeffbaumes:
Yeah good question, I do not have a great answer there.

The most user-efficient UI I can think of is a checkbox that states "I have confirmed these are the only features present" or "I have confirmed there are no features present" if they have selected none. If they've checked it, you submit everything as 1's or 0's, and if not you only submit the 1's and rejected ones as 0's.

Another option is to just always have all the chips in the field. More likely ones are first and say " - probable" and the rest are just " - unknown". Clicking once makes it present (1 value), clicking again makes it marked absent (0 value). But there may not be UI space for that. This is much like @curtislisle suggestion, but just has them all in one field but displayed differently.

dzenanz:
> always have all the chips in the field. More likely ones are first and say " - probable" and the rest are just " - unknown". Clicking once makes it present (1 value), clicking again makes it marked absent (0 value).

I like this the best. Have each chip have 3 states (preferably visually coded), and clicking them cycles through the states (present->absent->unknown->present->...).

Possible visual coding:
 * present = dark red filling with light (white?) font
 * absent = dark (black?) font, light green filling
 * unknown, unevaluated = gray font, white chip filling with dashed green-red outline

The mouse-over hint might contain the explanation of the visual coding.

curtislisle:
This is a nice approach.  Just to clarify, the NN will populate all the chips according to the calculated values and these values will be saved to the scan when the user clicks any scan decision: Ususable, unusable, or questionable.  However, the user can override anything they want, including setting  any values to be “undefined”, before they enter a scan decision.  Am I interpreting this right?

jeffbaumes:
@curtislisle I think you are close. As I understand the NN values will be in the database already before submitting any decision, so those don't need to be saved. So you'll have a set of floating-point NN predictions, as well as -1, 0, 1 user values. My main ask is to never just populate a 0/1 based on NN alone; the user needs to perform some action to validate. Chips can be "suggested" (my "probable" text idea) but not confirmed unless it is clicked.

@dzenanz I like the idea of cycling through all three states on click, that makes sense.

dzenanz:
Curtis, you are interpreting my idea exactly right.

> the user needs to perform some action to validate

Yes. Hopefully, they will just agree with NN's opinion a lot of the time. If they don't, they need to do some clicks on the chips before marking the scan as usable, unusable, or questionable.

And we will have some thresholds for converting [0.0, 1.0] score into the present/absent/unknown opinions. Maybe 25% or less is present, 25%-75% unknown, 75% and more is absent. Something like this. Perhaps establish the thresholds using some numerical analysis. Maybe even have 10 different thresholds for 10 different artifacts.

annehaley:
Okay, I like these ideas. How about a version where the artifacts show as chips and the user will toggle them through states of `confirmed_present`, `confirmed_absent`, and `unconfirmed`? An `unconfirmed` chip can be either `suggested` or `unsuggested`. The suggestion is determined by if a previous ScanDecision on the Scan has already applied some tags (then we suggest those) or else if there is some automatic Evaluation on the Scan (then we suggest those with values over the threshold of suggestion).

We can easily change the styling for each of these states. The current styling is like this:

1. Confirmed present = filled green
2. Confirmed absent = crossthrough text, unfilled
3. Unconfirmed suggested = green text, unfilled, followed by a "?"
4. Unconfirmed unsuggested = unfilled, normal text

The tooltip currently says:
> Toggle each tag below. Fill green to confirm an artifact is present.  Crossthrough to confirm an artifact is absent.


image

dzenanz:
Artifacts are bad. It is confusing to associate green color with their presence. I greatly prefer: green=good (artifact absent), red=bad (artifact present).

You suggest each chip to have 4 states. That incurs greater mental complexity than 3 states, and we are not using the distinction between the two suggested states - they both get recorded as `-1`. Also, cycling through 4 states requires more clicks than cycling through 3 states.

annehaley:
I agree that green for present is confusing when the artifacts have negative connotation, but I wasn't sure if red for present was appropriate either. Perhaps we could choose a styling independent of the red and green connotations?

The user does not cycle through four states, though. They cycle through unconfirmed -> confirmed present -> confirmed absent. The suggestion part is simply to determine the class that is used in the unconfirmed state. This is just to highlight the likely artifacts to the user when they are unconfirmed in either direction.

I think our main problem will be finding appropriate styling that can convey as much of this logic as we can.

annehaley:
I pushed a version that implements this idea. If anyone is able to check this out and play around with these buttons (try to break it  and feel out what might be good styling for the chip states), I'd appreciate your advice.

curtislisle:
I will try to pull down this version in the next few days and test the buttons.  Thanks for incorporating these ideas into working code already.

client/src/main.ts
```diff
@@ -45,14 +45,16 @@ Vue.config.productionTip = true;
 
 djangoRest.restoreLogin(store).then(async () => {
   await store.dispatch.loadMe();
+  await store.dispatch.loadConfiguration();
   const user = store.state.me;
+  const { MIQAConfig } = store.state;
```
dchiquito, line 50: inject/provide and the vuex are (loosely) two solutions to the same problem. Since it's available both from the store and via injection, there's some ambiguity as to where the "correct" place to get `MIQAConfig` from is. We could either just call `djangoRest` directly here and provide it, or not provide it and use `mapState` to get it from the store whenever it's needed.
(the same argument applies to `user`, that's probably a bigger refactor though)

This is just a gut feeling, I'm not that familiar with the pros/cons. Any thoughts @zachmullen ?

client/src/components/DecisionButtons.vue
```diff
@@ -16,13 +21,45 @@ export default {
...
+        return Object.entries(lastDecisionArtifacts).filter(
+          ([, present]) => present === 1,
+        ).map(([artifactName]) => artifactName);
+      } if (this.currentViewData.currentAutoEvaluation) {
+        const predictedArtifacts = this.currentViewData.currentAutoEvaluation.results;
+        return Object.entries(predictedArtifacts).filter(
+          ([artifactName, percentCertainty]) => artifactName !== 'overall_quality'
+            && artifactName !== 'normal_variants'
+            && percentCertainty < this.MIQAConfig.auto_artifact_threshold,
+        ).map(([artifactName]) => artifactName.replace('no_', '').replace('full', 'partial'));
```
dchiquito, line 59: I have absolutely no idea what this magic is. Can you add a little documentation for posterity?

client/src/django.ts
```diff
@@ -97,8 +101,15 @@ const djangoClient = {
     const { data } = await apiClient.get(`/scans/${scanId}`);
     return data;
   },
-  async setDecision(scanId: string, decision: string, comment: string) {
-    const { data } = await apiClient.post('/scan-decisions', { scan: scanId, decision, note: comment });
+  async setDecision(
+    scanId: string,
+    decision: string,
+    comment: string,
+    userIdentifiedArtifacts: Object,
```
dchiquito, line 108:
```suggestion
    userIdentifiedArtifacts: { present: Artifact[], absent: Artifact[] },
```
Artifacts are thrown around a lot, I think they'd benefit from having a well defined type.

miqa/core/rest/scan_decision.py
```diff
@@ -50,6 +51,22 @@ def create(self, request, **kwargs):
...
+            and 'present' in request_data['artifacts']
+            and 'absent' in request_data['artifacts']
+        ):
+            request_data['user_identified_artifacts'] = {
+                artifact_name: (
+                    1
+                    if artifact_name in request_data['artifacts']['present']
+                    else 0
+                    if artifact_name in request_data['artifacts']['absent']
+                    else -1
```
dchiquito, line 65: These (-1, 0, 1) values clearly have some semantic meaning, but I have to assume it's 1=present, -1=absent, 0=not set. An enum or some constants would make reading this code easier, and using a string-based enum (`"PRESENT", "ABSENT", "UNDEFINED") would make reading the data/frontend code easier as well.

client/src/components/DecisionButtons.vue
```diff
@@ -16,13 +21,45 @@ export default {
...
+    chips() {
+      return this.artifacts.map((artifact) => [artifact, this.getCurrentChipState(artifact)]);
+    },
+    suggestedArtifacts() {
+      if (this.currentViewData.scanDecisions.length > 0) {
+        const lastDecisionArtifacts = _.last(_.sortBy(
+          this.currentViewData.scanDecisions, (dec) => dec.created,
+        )).user_identified_artifacts;
+        return Object.entries(lastDecisionArtifacts).filter(
+          ([, present]) => present === 1,
```
dchiquito, line 51: a string enum on the backend would translate to a TS type with an enum field, which would remove this magic number.

dchiquito (changes requested):
Aesthetically looks good to me in terms of usage, just some implementation quibbles

client/src/main.ts
annehaley: My reasoning for this was the idea that we need to call `loadConfiguration` somewhere before render, just like we do for the user. It can be done in the `mounted` function, though, if we want.

client/src/django.ts
annehaley: The objects passed around are actually just strings, and I don't think they need more information, but we can absolutely replace `Object` here with `{ present: String[], absent: String[] }`.

miqa/core/rest/other_endpoints.py
```diff
@@ -0,0 +1,18 @@
...
+
+from miqa.core.models.scan_decision import default_identified_artifacts
+
+
+class MIQAConfigView(APIView):
+    @action(detail=True, methods=['GET'], permission_classes=[IsAuthenticated])
+    def get(self, request):
+        return Response(
+            {
+                'version': '1.0',
```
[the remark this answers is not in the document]
annehaley: Adding suggestion to batch of changes.

client/src/components/DecisionButtons.vue
annehaley: Added explanation comments.

miqa/core/rest/scan_decision.py
annehaley: Good idea!

client/src/main.ts
annehaley: We discussed this in a team meeting and decided that keeping this information as an injected variable is an acceptable approach.

PR #246, opened 2021-12-06
Admin account approval

### LINKED ISSUE #96: Require admin approval for new users
Right now in production anyone can sign up for an account. We should enable the admin approval flow so that users are not enabled until an admin approves them.

annehaley:
Resolves #96 

This PR implements the following flow:

1. User signs up -> user receives verification email
2. User verifies email -> user is reset to inactive and a notification email goes to all admin users.
3. User tries to log in -> user sees page explaining that an admin needs to approve their account
4. Any admin clicks the link in the email -> Admin chooses to deny with rejection reason
5. User is kept inactive and is notified that their account was rejected

OR
4. Any admin clicks the link in the email -> Admin chooses to approve the account
5. User is switched to active and is notified that they may now log in.

Since the approval link is sent to all admin users, any admin may override the decision of another. The last admin's decision is used, but the new user will be notified of every decision made.

### DIFF
--- a/miqa/core/rest/accounts.py
+++ b/miqa/core/rest/accounts.py
@@ -15,7 +15,7 @@ class AccountInactiveView(TemplateView):
 
 class AccountActivateView(TemplateView):
     def dispatch(self, request, *args, **kwargs):
-        if not request.user.is_authenticated and request.user.is_superuser:
+        if not (request.user.is_authenticated and request.user.is_superuser):
             return redirect('home')
         return super(AccountActivateView, self).dispatch(request, *args, **kwargs)
 
--- a/miqa/core/rest/accounts.py
+++ b/miqa/core/rest/accounts.py
@@ -1,4 +1,9 @@
+from django.conf import settings
 from django.contrib.auth import logout
+from django.contrib.auth.models import User
+from django.core.exceptions import PermissionDenied
+from django.core.mail import EmailMultiAlternatives
+from django.shortcuts import render
 from django.views.generic.base import TemplateView
 from rest_framework import status
 from rest_framework.response import Response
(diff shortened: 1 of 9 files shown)


miqa/core/rest/accounts.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class AccountInactiveView(TemplateView):
+    template_name = 'account_inactive.html'
+
+
+class AccountActivateView(TemplateView):
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.is_authenticated and request.user.is_superuser:
+            return redirect('home')
```
dchiquito, line 19: This isn't working in my production deployment. I was somehow able to approve myself without logging in at all.
```suggestion
        if not (request.user.is_authenticated and request.user.is_superuser):
            return redirect('home')
```
Perhaps python was parsing it as `(not authenticated) and superuser: go(home)`

miqa/core/signals.py
```diff
@@ -0,0 +1,35 @@
...
+
+    msg = EmailMultiAlternatives(
+        'MIQA Account Approval - New User',
+        f'A new user with the email {kwargs["email_address"]} has created '
+        'an account in MIQA. As an administrative user, it is your responsibility '
+        'to activate this account if you believe this user is legitimate. '
+        'If you believe this account should not be activated, reach out to this user'
+        f'{" and other administrators." if len(admins) > 1 else "."} '
+        '\n\n '
+        f'To activate this account, visit {activation_link}. '
```
dchiquito, line 29:
```suggestion
        f'To activate this account, visit {activation_link} '
```
For easier copy/pasting.

Ideally we should have an HTML email body as well as a plaintext so that links can be rendered.

dchiquito (dismissed):

PR #247, opened 2021-12-06
Use TorchIO for corrupting images with synthetic ghosting and motion

### LINKED ISSUE #27: Utilize synthetic bad images
https://torchio.readthedocs.io/transforms/augmentation.html

dzenanz:
This adds augmentation of bad images during training phase. Inference remains unaffected.

Closes #27.


annehaley (approved):
I verified that the inference process is unaffected, but I'm afraid I can't verify much else. The code looks good, though!

PR #256, opened 2021-12-09
Clearer import errors

### LINKED ISSUE #182: Frontend should display more descriptive error messages
For example, when an Import fails, the snackbar notification displays "Import filed. See server logs for details." This does not tell the user if the failure is due to something they can solve, like being unauthenticated or specifying the import path wrong.

The server failure responses should include the error message that the client should display.

### LINKED ISSUE #251: CSV with header but no entries results in breaking MIQA and showing unreadable error
This is related to Girder-3 version but might happen in Girder-4 as well

annehaley:
Resolves #182.
This might resolve #251; I couldn't replicate the issue with an unpopulated import file, but I did notice that the import error messages were always vague. This PR addresses that by propagating server-side errors to the snackbar error message display in the frontend.

### DIFF
--- a/miqa/core/rest/project.py
+++ b/miqa/core/rest/project.py
@@ -2,6 +2,7 @@ from drf_yasg.utils import no_body, swagger_auto_schema
 from guardian.shortcuts import get_objects_for_user, get_users_with_perms
 from rest_framework import serializers, status
 from rest_framework.decorators import action
+from rest_framework.exceptions import APIException
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.viewsets import ReadOnlyModelViewSet
@@ -137,7 +138,7 @@ class ProjectViewSet(ReadOnlyModelViewSet):
             # tasks sent to celery must use serializable arguments
             import_data(request.user.id, project.id)
         except Exception as e:
-            return Response({'message': str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+            raise APIException(str(e))
 
         return Response(status=status.HTTP_204_NO_CONTENT)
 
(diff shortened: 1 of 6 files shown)


dchiquito (changes requested):

miqa/core/rest/project.py
```diff
@@ -133,8 +134,11 @@ def settings_(self, request, **kwargs):
     def import_(self, request, **kwargs):
         project: Project = self.get_object()
 
-        # tasks sent to celery must use serializable arguments
-        import_data(request.user.id, project.id)
+        try:
+            # tasks sent to celery must use serializable arguments
+            import_data(request.user.id, project.id)
+        except Exception as e:
+            raise APIException(str(e))
```
dchiquito, line 141:
```suggestion
        # tasks sent to celery must use serializable arguments
        import_data(request.user.id, project.id)
```
IMHO the benefit of using `APIException`s is that you can throw them when you encounter an error anywhere in your code, and DRF will intelligently convert them into a response code. In this case you would add them to `import_data`, something like this:
```
    if project.import_path.endswith('.csv'):
        import_dict = import_dataframe_to_dict(pandas.read_csv(project.import_path))
    elif project.import_path.endswith('.json'):
        import_dict = json.load(open(project.import_path))
    else:
        raise APIException(f'Invalid import file {project.import_path}.', code=status.HTTP_400_BAD_REQUEST)
```
This will raise a 400 error with an `Invalid import file ...` detail field. (the default code 500 denotes an internal server error, which doesn't seem appropriate here since the user set an invalid import_path).

dchiquito (approved):
Raw `APIException`s result in a 500 error which isn't appropriate for most of these failure modes, but nothing is consuming the status code, so I think it's fine for now. We would have to either specify the code explicitly in every `APIException` instantiation, or define our own `class ImportException(APIException)` with an appropriate code.

annehaley:
> We would have to either specify the code explicitly in every `APIException` instantiation, or define our own `class ImportException(APIException)` with an appropriate code.

Should I file an issue for this? Or do it now?

dchiquito:
This is blocking another PR so I'd say just file an issue for later.

PR #265, opened 2021-12-29
Fix crosshairs

annehaley:
After the changes made to `VtkViewer` to fix the camera placement for flipped/rotated/axis-switched images, the assumptions made for drawing crosshairs were no longer correct. This PR changes the crosshairs logic to account for this.

### DIFF
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -198,9 +198,11 @@ export default {
     trueAxis(axisName) {
       const orientation = this.representation.getInputDataSet().getDirection();
       const axisNumber = VIEW_ORIENTATIONS[axisName].axis;
-      const axisOrientation = orientation.slice(
-        axisNumber * 3, axisNumber * 3 + 3,
-      ).map(
+      const axisOrientation = [
+        orientation[axisNumber],
+        orientation[3 + axisNumber],
+        orientation[6 + axisNumber],
+      ].map(
         (val) => Math.abs(val),
       );
       const axisOrdering = ['x', 'y', 'z'];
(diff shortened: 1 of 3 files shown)


client/src/components/VtkViewer.vue
```diff
@@ -195,54 +195,19 @@ export default {
...
+      const orientation = this.representation.getInputDataSet().getDirection();
+      const axisNumber = VIEW_ORIENTATIONS[axisName].axis;
+      const axisOrientation = orientation.slice(
+        axisNumber * 3, axisNumber * 3 + 3,
+      ).map(
+        (val) => Math.abs(val),
+      );
+      const axisOrdering = ['x', 'y', 'z'];
+      const trueAxis = axisOrdering[
+        axisOrientation.indexOf(Math.max(...axisOrientation))
```
dzenanz, line 210: Wouldn't this select some axis twice in certain oblique cases? What does `oblique2` test case look like now? Will this interact with the expected fix of https://github.com/Kitware/vtk-js/issues/2191?

dzenanz (commented):
Looks like a great simplification!

client/src/components/VtkViewer.vue
annehaley: Good point. It may not be a problem once that bug is fixed, but it wouldn't hurt to come up with a more foolproof way of selecting the true axis. I'm not sure the best way. What do you recommend?

I recorded the current behavior of CTChestOblique2rot in the following video:

https://user-images.githubusercontent.com/44912689/147777237-94a4e920-325e-481e-9c63-a1948099e8f1.mov

dzenanz: I guess it was like this before this simplification? As long as we have the more complicated code in history or in a branch, we should move forward with this. When vtk.js bug gets fixed, we should revisit this. Then either revive or totally discard the more complicated code.

PR #268, opened 2022-01-04
Use npm instead of yarn in CI

zachmullen:
Also use node 16


dzenanz (approved):

client/package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "vue-cli-service lint"
   },
   "dependencies": {
-    "@girder/components": "git+https://github.com/girder/girder_web_components.git#v3.0.2",
+    "@girder/components": "3.0.2",
```
zachmullen, line 11: @dchiquito PTAL, was there some reason we were pulling from git instead of npm registry?

annehaley (approved):
LGTM

client/package.json
dchiquito: no idea ¯\_(ツ)_/¯

PR #269, opened 2022-01-04
Jettison GWC

zachmullen:
As far as I can tell, we no longer need GWC. We were only using it for the now unused LoginView, and to configure mdi fonts as a side effect.


annehaley (approved):
This makes sense to me. I had been wondering why `Login.vue` was around.

PR #271, opened 2022-01-04
Improve vtk viewer toolbar positioning

zachmullen:
It now no longer overflows the container. Before:

Screen Shot 2022-01-04 at 1 46 21 PM

After:

Screen Shot 2022-01-04 at 1 45 16 PM

### DIFF
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -364,8 +364,9 @@ export default {
       class="toolbar"
       dark
       flat
-      color="black"
-      max-height="42"
+      color="#ffffff00"
+      height="46"
+      max-height="46"
     >
       <div
         :class="name"


client/src/components/VtkViewer.vue
```diff
@@ -364,8 +364,9 @@ export default {
       class="toolbar"
       dark
       flat
-      color="black"
-      max-height="42"
+      color="#ffffff00"
```
zachmullen, line 367: I may be over-interpreting here; I'm assuming our intention is that the toolbar should be the same color as the container it's inside of. If that's not the case, we can back out the color change here.

annehaley: That's a good point and I like this better.

annehaley (approved):

PR #276, opened 2022-01-06
Project page design changes

### LINKED ISSUE #273: Changes to project page design
In our meeting with Stanford, we decided to make the following revisions to the project page design:
- color-code the bullets next to scan names according to the scan state (matching the donut colors) and add a tooltip to show the state as text
- Add an overall status next to project names in the project list, i.e. text like "(2/9)" with a tooltip like "2 of 9 scans complete"

annehaley:
The only portion I had trouble with was defining the colors of the bullet points in `ExperimentsList`. It appears that I can't dynamically apply the color to the `marker` pseudo element, so I had to copy paste the hex values defined in the `ScanState` enum in `types.ts`. Is there a way I can insert these values dynamically instead of having these values defined in two places?


client/src/components/ExperimentsView.vue
```diff
@@ -114,26 +127,32 @@ export default {
             <li
               v-for="scan of scansForExperiment(experiment.id)"
               :key="`s.${scan.id}`"
-              :class="{
-                current: scan === currentScan
-              }"
-              class="body-1"
+              :class="'body-1 state-'+scanState(scan).replace(/ /g, '-')+scanIsCurrent(scan)"
```
dchiquito, line 130: I like to minimize the amount of logic in the templates, so it seems cleaner to use a `scanStateClass` method that does something like this:
```
scanStateClass(scan) {
  var classes = `body-1 state-{this.scanState(scan).replace(/ /g, '-')}`;
  if (scan === this.currentScan) {
    classes = `{classes} current`;
  }
  return classes;
}
```

miqa/core/models/project.py
```diff
@@ -77,6 +77,26 @@ def get_user_role(self, user):
...
+                    if scan.decisions.count() > 0
+                    and scan.decisions.latest('created').creator.id in tier_2_reviewers
+                ]
+            )
+            for exp in self.experiments.all()
+        ]
+        return {
+            'total_scans': sum([exp.scans.count() for exp in self.experiments.all()]),
+            'total_complete': sum(complete_scans_per_exp),
+        }
```
dchiquito, line 98: Does this need to be added to the TS Project type?

dchiquito (commented):
Regarding the CSS thing, I think that's a problem that is generally solved with SCSS/Sass. I have no idea what the best practices are for that though. @zachmullen ?

miqa/core/models/project.py
annehaley: Good point. Added.

zachmullen: The above block looks like it may contain n+1 query issues. Can you check out the SQL queries generated by this function and verify one way or the other?

annehaley: I've admittedly never tried looking directly at SQL queries from django QuerySets. How can I dig down to see that?

zachmullen: The django-debug-toolbar is the way to do it for server rendered pages, but since we're using an SPA, that won't work. There are a couple ways to do it that I like to do. If it's straightforward, you can use `manage.py shell_plus --print-sql`, and in the shell, call this function. If for some reason it's not practical to invoke this method in the shell, you can use `manage.py runserver_plus --print-sql` (which involves changing your docker-compose.override.yml if you're using docker-compose).

annehaley: Thanks, I'll try it out!

annehaley: My previous attempt made far too many database hits, and it was slowing down the response time. In dfcec53, I was able to make use of `prefetch_related` to reduce this effect.

client/src/components/ExperimentsView.vue
```diff
@@ -76,6 +77,25 @@ export default {
...
+    scanState(scan) {
+      if (!this.currentTaskOverview) {
+        return 'unreviewed';
+      }
+      return this.currentTaskOverview.scan_states[scan.id];
+    },
+    scanStateClass(scan) {
+      let classes = `body-1 state-${this.scanState(scan).replace(/ /g, '-')}`;
+      if (scan === this.currentScan) {
+        classes = '{classes} current';
```
dchiquito, line 95:
```suggestion
        classes = `${classes} current`;
```

miqa/core/models/project.py
```diff
@@ -77,6 +77,24 @@ def get_user_role(self, user):
...
+        complete_scans_in_project = [
+            scan
+            for scan in scans_in_project
+            if scan.decisions.count() > 0
+            and scan.decisions.latest('created').creator.id in tier_2_reviewers
+        ]
+        return {
+            'total_scans': scans_in_project.count(),
+            'total_complete': len(complete_scans_in_project),
+        }
```
zachmullen, line 96: Much better, thanks!

miqa/core/models/project.py
```diff
@@ -77,6 +77,24 @@ def get_user_role(self, user):
...
+        ]
+        scans_in_project = Scan.objects.filter(
+            experiment__in=self.experiments.all()
+        ).prefetch_related('decisions')
+        complete_scans_in_project = [
+            scan
+            for scan in scans_in_project
+            if scan.decisions.count() > 0
+            and scan.decisions.latest('created').creator.id in tier_2_reviewers
+        ]
```
zachmullen, line 92: FWIW, this filter could also be expressed as part of the initial query, but I don't want that to hold up this PR. It would be a good learning exercise, though!

dchiquito (approved):

PR #280, opened 2022-01-10
Enable navigating directly to Frame

### LINKED ISSUE #233: Allow navigating directly to a frame
In the original Girder 3 app, you could navigate directly to a frame using a URL like `/#/b0c07a68-11a5-4600-93d3-666aa7381e6c`. Right now, we are navigating to the projects page if the current project isn't defined, which is only the case when refreshing/navigating directly: https://github.com/OpenImaging/miqa/blob/master/client/src/views/Frame.vue#L55-L59

We should either load the project corresponding to the frame, or include the project ID in the URL fragment (i.e. `/#/{project_uuid}/{frame_uuid}`.

annehaley:
Resolves #233. I went with @dchiquito's suggestion to include the project id in the url.

I left the following line in just in case there is some failure to load the project:
```
this.$router.replace('/').catch(this.handleNavigationError);
```
But, in most cases, this line is not reached.

### DIFF
--- a/client/src/store/index.ts
+++ b/client/src/store/index.ts
@@ -347,6 +347,7 @@ const {
         downTo = null;
       }
       return {
+        projectId: project.id,
         projectName: project.name,
         experimentId: experiment.id,
         experimentName: experiment.name,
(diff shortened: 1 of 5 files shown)


client/src/store/index.ts
```diff
@@ -686,6 +679,16 @@ const {
         },
       });
     },
+    async getFrame({ state, dispatch }, { frameId, projectId }) {
+      if (!frameId) {
+        return undefined;
+      } if (!state.frames[frameId]) {
```
dchiquito, line 685: I'm surprised this passes linting
```suggestion
      }
      if (!state.frames[frameId]) {
```

dchiquito (approved):

PR #282, opened 2022-01-14
Cancel pending XHRs on experiment switch, and add better progress UX on frame download

### LINKED ISSUE #255: Present state of pre-fetching for each scan in the UI

zachmullen:
Fixes #255 

Screenshot of progress when awaiting frame download:

Screen Shot 2022-01-14 at 10 07 03 AM


annehaley (approved):
This makes sense to me. I wasn't able to test it with very large datasets, so I only ever saw the loading circle at 100%, but I can assume it works based on reading this.

zachmullen:
It's a good practice to occasionally go through app workflows under slow network conditions to try out the user experience. Luckily, Chrome and Firefox provide great devtools to do that. In the developer console, go to the Network tab and you'll see a dropdown that says "No throttling". You can set that to a slow connection (I used "Fast 3G" or "Slow 3G" to test out this feature). It's very handy!

annehaley:
Very cool! Just tried it and the loader looks great!

PR #288, opened 2022-01-19
Add support for nrrd and mgz file types

### LINKED ISSUE #262: Support .nrrd files
MIQA currently will not render `nrrd` files at all, even if the content is exactly the same as a `nifti` file that it can render.

annehaley:
Resolves #262 

There were some places in the frontend that we used a magic filename `nifti.nii.gz` passed into the file reader, so I had to add a bit to the Frame serializer on the backend to tell us what the true file extension was.


dchiquito (approved):
Code looks good to me, but I defer to @curtislisle

Issue #274, opened 2022-01-05
Redesign global imports


annehaley:
In our recent meeting with Stanford, we discussed possible new implementations for the global import feature. @dchiquito and I decided on a satisficing solution. Tasks to accomplish this are as follows:

- [ ] Replace `global_import_export` boolean field on Project to `is_global` and add a constraint that only one existing Project can be global. 
- [ ] Add an endpoint `get_global_project` that returns the id of the Project for which `is_global` is True
- [ ] Remove the "global import/export" toggle from the `ProjectSettings` component
- [ ] Add a button "Global Import/Export" to the bottom of the sidebar on the Projects page. This button should get the global project and display a similar page as other Projects, except it only has the first card, `ProjectSettings`.
- [ ] Modify the logic in the import endpoint. If the target Project has `is_global` as False, perform a normal import with that Project's`import_csv` that ignores the `project_name` column and puts all rows in the current Project. If the target project has `is_global` as True, perform a global import with that Project's csv, potentially modifying existing Projects and creating new ones.

zachmullen:
I'm going to make a PR showing an alternate implementation on the backend.

annehaley:
Resolved with #292

PR #292, opened 2022-01-20
Global settings singleton model

zachmullen:
This shows my proposed design for a singleton model that could hold all of our global configuration values that need to be editable at runtime. Usage:

```python
gs = GlobalSettings.load()
print(gs.import_path)  # access the global import path
gs.import_path = '/new/import/path'
gs.save(update_fields=['import_path'])  # update the global import path
```

If this is approved, we have to consider how the REST endpoints would look. I think the best way would be to have `GET` and `PUT` endpoints for the global settings values, and then global `import` and `export` endpoints that don't live under the `project` namespace.

Refs #274


miqa/core/models/global_settings.py
```diff
@@ -0,0 +1,16 @@
...
+    import_path = models.CharField(max_length=1000, default='', blank=True)
+    export_path = models.CharField(max_length=1000, default='', blank=True)
+
+    def save(self, *args, **kwargs):
+        self.pk = 1
+        super().save(*args, **kwargs)
+
+    @classmethod
+    def load(cls):
+        return cls.objects.get_or_create(pk=1)[0]
```
zachmullen, line 16: Note that these `save` and `load` methods are all that's needed for the singleton behavior.

annehaley (approved):
LGTM, is the next step taking this to the frontend?

zachmullen:
Before I merge this, I'll add something to the migration to convert from the global project to the new settings, then delete the global project. After that, we'll do the REST updates, and then finally the client changes.

zachmullen:
I'm done with the part I wanted to do. If we approve this, my suggestion would be to make other PRs against this branch, then merge all at once.

dchiquito (approved):

annehaley:
Now that the frontend changes are shown here, perhaps a squash and merge would be best going into `master`.

client/src/store/index.ts
```diff
@@ -283,7 +283,8 @@ const initState = {
   me: null,
   allUsers: [],
   reviewMode: false,
-  currentProject: null as Project | null,
+  globalSettings: undefined as ProjectSettings,
+  currentProject: undefined as Project | null,
```
dchiquito, line 287: For my own understanding, `currentProject == undefined` means no project is selected, `currentProject == null` means the global project is selected, otherwise `currentProject` is the selected project?

miqa/core/migrations/0025_globalsettings.py
```diff
@@ -0,0 +1,36 @@
...
+from django.db import migrations, models
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+from django.db.migrations.state import StateApps
+
+
+def migrate_global_project(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor):
+    Project = apps.get_model('core', 'Project')  # noqa: N806
+    GlobalSettings = apps.get_model('core', 'GlobalSettings')  # noqa: N806
+
+    if global_project := Project.objects.filter(name='global').first():
```
dchiquito, line 12:
```suggestion
    if global_project := Project.objects.filter(global_import_export=True).first():
```

miqa/core/migrations/0025_globalsettings.py
```diff
@@ -0,0 +1,36 @@
...
+        migrations.CreateModel(
+            name='GlobalSettings',
+            fields=[
+                ('id', models.PositiveSmallIntegerField(primary_key=True, serialize=False)),
+                ('import_path', models.CharField(blank=True, default='', max_length=1000)),
+                ('export_path', models.CharField(blank=True, default='', max_length=1000)),
+            ],
+        ),
+        migrations.RunPython(migrate_global_project),
+    ]
```
dchiquito, line 36: I think you also need to remove `global_import_export` from the `Project` model

miqa/core/conversion/import_export_csvs.py
```diff
@@ -22,8 +23,13 @@ def validate_file_locations(input_dict, project):
         if key == 'file_location':
             raw_path = Path(value)
             if not raw_path.is_absolute():
-                # not an absolute file path; refer to project import csv location
-                raw_path = Path(project.import_path).parent.parent / raw_path
+                import_path = (
+                    GlobalSettings.load().import_path
+                    if project == 'global'
```
dchiquito, line 28: This appears to be the only place where the magic string `global` is still used. Is this left over from something?

dchiquito (changes requested):

client/src/store/index.ts
annehaley: Correct.

client/src/components/DataImportExport.vue
```diff
@@ -67,9 +79,11 @@ export default defineComponent({
       }
       exporting.value = false;
     }
+    const { isGlobal } = store.getters;
```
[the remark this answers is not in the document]
annehaley: Simply for reactivity; when this const was set with the rest of the `refs`, it didn't update when `currentProject` changed. This was the first solution I could think of. Do you know of a more acceptable way to make this getter reactive to changes to `state.currentProject`?

miqa/core/conversion/import_export_csvs.py
annehaley: Yes, nice catch. changed in [e329dcd](https://github.com/OpenImaging/miqa/pull/292/commits/e329dcd50dbc882992b6d7016f8d982222d87ffa)

miqa/core/migrations/0025_globalsettings.py
annehaley: Good point. And with the new migration, I figured we could use a squash as well. changed in [b05b85c](https://github.com/OpenImaging/miqa/pull/292/commits/b05b85ca53839899deb48d13b37bca03d683b782)

miqa/core/migrations/0025_globalsettings.py
annehaley: changed in [b05b85c](https://github.com/OpenImaging/miqa/pull/292/commits/b05b85ca53839899deb48d13b37bca03d683b782)

client/src/components/DataImportExport.vue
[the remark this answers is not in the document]
dchiquito:
```js
const isGlobal = computed(() => store.getters.isGlobal);
```
might work, not sure though. If it doesn't then what you have here is fine.

PR #293, opened 2022-01-25
Fix Zoom bug

### LINKED ISSUE #277: Image zoom resets by itself
Let's say the user has zoomed into one or more of MIQAs images, by using the right mouse button and dragging, during a review and then their mouse cursor accidentally brushes over the left side of the browser (showing the Experiment list, even if briefly).  The zooms of all axes views are restored to their original starting position.  If it isn't too hard to do, the zoom level of the images should be preserved unless the user refreshes the browser page, in my opinion.

annehaley:
Resolves #277.

There are two approaches in the two below commits. The first introduces a logic in the `onWindowResize` method such that `fill2DView` is not called if the zoom level is manually set by the user. However, it may be simpler to remove this definition of the `onWindowResize` from `VtkViewer` entirely. It is unclear the original purpose for this function, because without it expected behavior is still seen. 

Could I get help determining whether it really is the case that we don't need that function anymore?

### DIFF
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -25,6 +25,7 @@ export default {
     resized: false,
     fullscreen: false,
     screenshotContainer: document.createElement('div'),
+    originalZoom: undefined,
   }),
   computed: {
     ...mapState(['proxyManager', 'loadingFrame', 'showCrosshairs', 'iIndexSlice', 'jIndexSlice', 'kIndexSlice']),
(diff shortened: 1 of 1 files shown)


dchiquito (approved):
It seems like resizing the window still resets the zoom level, but mousing over the nav menu no longer does, which is all that's required to fix this zoom bug. I can't find anything that this change breaks, so even though I'm not sure what that watcher was originally for, I think removing it is a good idea.

PR #294, opened 2022-01-26
Global settings frontend

annehaley:
This PR introduces four new endpoints to accompany the `GlobalSettings` singleton model:

1. GET /global/settings
2. PUT /global/settings
3. POST /global/import
4. POST /global/export

Additionally, this PR leverages those endpoints on the frontend to include an admin-only UI for editing the values of th global import path and export path, as well as performing global imports/exports

### DIFF
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -20,6 +20,7 @@ export default defineComponent({
   setup() {
     const currentProject = computed(() => store.state.currentProject);
     const loadProject = (project: Project) => store.dispatch.loadProject(project);
+    const { isGlobal } = store.getters;
 
     const importing = ref(false);
     const importDialog = ref(false);
@@ -32,7 +33,11 @@ export default defineComponent({
       importErrorText.value = '';
       importErrors.value = false;
       try {
-        await djangoRest.import(currentProject.value.id);
+        if (isGlobal) {
+          await djangoRest.globalImport();
+        } else {
+          await djangoRest.projectImport(currentProject.value.id);
+        }
         importing.value = false;
 
         this.$snackbar({
@@ -40,7 +45,7 @@ export default defineComponent({
           timeout: 6000,
         });
 
-        if (currentProject.value.id !== 'global') {
+        if (!isGlobal) {
           await loadProject(currentProject.value);
         }
       } catch (ex) {
@@ -55,7 +60,11 @@ export default defineComponent({
     async function exportData() {
       exporting.value = true;
       try {
-        await djangoRest.export(currentProject.value.id);
+        if (isGlobal) {
+          await djangoRest.globalExport();
+        } else {
+          await djangoRest.projectExport(currentProject.value.id);
+        }
         this.$snackbar({
           text: 'Saved data to file successfully.',
           timeout: 6000,
@@ -72,6 +81,7 @@ export default defineComponent({
 
     return {
       currentProject,
+      isGlobal,
       loadProject,
       importing,
       importDialog,
@@ -144,7 +154,7 @@ export default defineComponent({
         <v-card-title class="title">
           Import
         </v-card-title>
-        <v-card-text v-if="currentProject.id == 'global'">
+        <v-card-text v-if="isGlobal">
           Importing data will overwrite all objects in every project listed in the import file.
           Are you sure you want to overwrite all objects in multiple projects?
         </v-card-text><v-card-text v-else>
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -20,7 +20,6 @@ export default defineComponent({
   setup() {
     const currentProject = computed(() => store.state.currentProject);
     const loadProject = (project: Project) => store.dispatch.loadProject(project);
-    const { isGlobal } = store.getters;
 
     const importing = ref(false);
     const importDialog = ref(false);
@@ -29,6 +28,7 @@ export default defineComponent({
     const exporting = ref(false);
 
     async function importData() {
+      const { isGlobal } = store.getters;
       importing.value = true;
       importErrorText.value = '';
       importErrors.value = false;
@@ -58,8 +58,11 @@ export default defineComponent({
       importDialog.value = false;
     }
     async function exportData() {
+      const { isGlobal } = store.getters;
       exporting.value = true;
       try {
+        console.log('global', isGlobal);
+
         if (isGlobal) {
           await djangoRest.globalExport();
         } else {
@@ -78,6 +81,7 @@ export default defineComponent({
       }
       exporting.value = false;
     }
+    const { isGlobal } = store.getters;
 
     return {
       currentProject,
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -61,8 +61,6 @@ export default defineComponent({
       const { isGlobal } = store.getters;
       exporting.value = true;
       try {
-        console.log('global', isGlobal);
-
         if (isGlobal) {
           await djangoRest.globalExport();
         } else {
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -40,7 +40,9 @@ export default defineComponent({
           timeout: 6000,
         });
 
-        await loadProject(currentProject.value);
+        if (currentProject.value.id !== 'global') {
+          await loadProject(currentProject.value);
+        }
       } catch (ex) {
         importing.value = false;
         this.$snackbar({
@@ -102,7 +104,7 @@ export default defineComponent({
           Import
         </v-btn>
       </template>
-      <span>Import from {{ currentProject.settings.importPath }}</span>
+      <span>Import from import path</span>
     </v-tooltip>
 
     <v-tooltip
@@ -130,7 +132,7 @@ export default defineComponent({
           </span>
         </v-btn>
       </template>
-      <span>Export to {{ currentProject.settings.exportPath }}</span>
+      <span>Export to export path</span>
     </v-tooltip>
 
     <v-dialog
@@ -142,7 +144,10 @@ export default defineComponent({
         <v-card-title class="title">
           Import
         </v-card-title>
-        <v-card-text>
+        <v-card-text v-if="currentProject.id == 'global'">
+          Importing data will overwrite all objects in every project listed in the import file.
+          Are you sure you want to overwrite all objects in multiple projects?
+        </v-card-text><v-card-text v-else>
           Importing data will overwrite all objects in this project, do you want
           to continue?
         </v-card-text>
--- a/client/src/django.ts
+++ b/client/src/django.ts
@@ -1,7 +1,7 @@
 import axios from 'axios';
 import OAuthClient from '@girder/oauth-client';
 import {
-  Project, ProjectTaskOverview, Settings, User,
+  Project, ProjectTaskOverview, ProjectSettings, User,
 } from './types';
 import { API_URL, OAUTH_API_ROOT, OAUTH_CLIENT_ID } from './constants';
 
@@ -50,11 +50,23 @@ const djangoClient = {
     const { data } = await apiClient.get('/configuration/');
     return data;
   },
+  async globalSettings() {
+    const { data } = await apiClient.get('/global/settings')
+    return data;
+  },
   async import(projectId: string) {
-    await apiClient.post(`/projects/${projectId}/import`);
+    if( projectId === 'global' ){
+      await apiClient.post(`/global/import`);
+    } else {
+      await apiClient.post(`/projects/${projectId}/import`);
+    }
   },
   async export(projectId: string) {
-    return apiClient.post(`/projects/${projectId}/export`);
+    if( projectId === 'global' ){
+      return apiClient.post(`/global/export`);
+    } else {
+      return apiClient.post(`/projects/${projectId}/export`);
+    }
   },
   async createProject(projectName: string): Promise<Project> {
     const { data } = await apiClient.post('/projects', { name: projectName });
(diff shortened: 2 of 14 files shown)


client/src/components/DataImportExport.vue
```diff
@@ -40,7 +40,9 @@ export default defineComponent({
           timeout: 6000,
         });
 
-        await loadProject(currentProject.value);
+        if (currentProject.value.id !== 'global') {
```
zachmullen, line 43: Would it be difficult to have global indicated by `currentProject === null` rather than this special `id` value?

client/src/components/DataImportExport.vue
```diff
@@ -142,7 +144,10 @@ export default defineComponent({
         <v-card-title class="title">
           Import
         </v-card-title>
-        <v-card-text>
+        <v-card-text v-if="currentProject.id == 'global'">
```
zachmullen, line 147: Given that we use this same condition in a multiple places, let's make `isGlobal` a getter in the Vuex store.

client/src/django.ts
```diff
@@ -50,10 +50,21 @@ const djangoClient = {
     const { data } = await apiClient.get('/configuration/');
     return data;
   },
+  async globalSettings() {
+    const { data } = await apiClient.get('/global/settings');
+    return data;
+  },
   async import(projectId: string) {
```
zachmullen, line 57: Might be best to split this up into `globalImport()` and `projectImport(projectId: string)` to make it more explicit for callers.

client/src/django.ts
```diff
@@ -77,11 +88,15 @@ const djangoClient = {
...
-  async settings(projectId: string): Promise<Settings> {
+  async settings(projectId: string): Promise<ProjectSettings> {
     const { data } = await apiClient.get(`/projects/${projectId}/settings`);
     return data;
   },
-  async setSettings(projectId: string, settings: Settings) {
+  async setSettings(projectId: string, settings: ProjectSettings) {
+    if (projectId === 'global') {
+      const resp = await apiClient.put('/global/settings', settings);
+      return resp.status === 200 ? resp.data : null;
```
zachmullen, line 98: Seems like we want an `else` after this, no? Or better yet, break it up into `setGlobalSettings` and `setProjectSettings`

client/src/views/Projects.vue
```diff
@@ -117,10 +118,13 @@ export default defineComponent({
   <div>
     <Navbar />
     <div class="d-flex">
-      <v-card>
+      <v-card style="height: calc(100vh - 50px)">
```
zachmullen, line 121: Should we move this style rule behind a CSS class?

miqa/core/conversion/import_export_csvs.py
```diff
@@ -33,7 +38,7 @@ def validate_file_locations(input_dict, project):
     return input_dict
 
 
-def validate_import_dict(import_dict, project: Project):
+def validate_import_dict(import_dict, project):
```
zachmullen, line 41: I'd suggest we change this to 
```suggestion
def validate_import_dict(import_dict, project: Optional[Project]):
```

And have global status indicated by `project is None`.

miqa/core/rest/global_settings.py
```diff
@@ -0,0 +1,74 @@
...
+        methods=['GET', 'PUT'],
+    )
+    def settings_(self, request):
+        if request.method == 'PUT':
+            global_settings = self.get_object()
+            global_settings.import_path = request.data['importPath']
+            global_settings.export_path = request.data['exportPath']
+            global_settings.full_clean()
+            global_settings.save()
+        return Response(GlobalSettingsSerializer(self.get_object()).data, status=status.HTTP_200_OK)
```
zachmullen, line 54:
```suggestion
        return Response(GlobalSettingsSerializer(self.get_object()).data)
```

miqa/core/rest/global_settings.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @swagger_auto_schema(responses={204: 'Import succeeded.'})
+    @action(
+        detail=False,
+        url_path='import',
+        methods=['POST'],
+    )
+    def import_(self, request, **kwargs):
+        import_data('global')
```
zachmullen, line 63:
```suggestion
        import_data(None)
```

miqa/core/rest/global_settings.py
```diff
@@ -0,0 +1,74 @@
...
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @swagger_auto_schema(responses={204: 'Export succeeded.'})
+    @action(
+        detail=False,
+        url_path='export',
+        methods=['POST'],
+    )
+    def export_(self, request, **kwargs):
+        export_data('global')
```
zachmullen, line 73:
```suggestion
        export_data(None)
```

miqa/core/tasks.py
```diff
@@ -44,15 +44,20 @@ def evaluate_data(frame_ids, project_id):
         )
 
 
-def import_data(user_id, project_id):
-    project = Project.objects.get(id=project_id)
+def import_data(project_id):
```
zachmullen, line 47:
```suggestion
def import_data(project_id: Optional[str]):
```

zachmullen (changes requested):

client/src/components/DataImportExport.vue
annehaley: changed in f932555

client/src/components/DataImportExport.vue
annehaley: changed in f932555

client/src/django.ts
annehaley: changed in f932555

client/src/django.ts
annehaley: changed in f932555

client/src/views/Projects.vue
annehaley: changed in f932555

miqa/core/conversion/import_export_csvs.py
annehaley: changed in f932555

miqa/core/rest/global_settings.py
annehaley: changed in f932555

miqa/core/rest/global_settings.py
annehaley: changed in f932555

miqa/core/rest/global_settings.py
annehaley: changed in f932555

miqa/core/tasks.py
annehaley: changed in f932555

client/src/store/index.ts
```diff
@@ -568,6 +571,21 @@ const {
...
+        name: null,
+        experiments: null,
+        settings: {
+          importPath: globalSettings.import_path,
+          exportPath: globalSettings.export_path,
+          permissions: null,
+        },
+        status: null,
+      };
+      commit('setCurrentProject', pseudoProject);
```
zachmullen, line 587: Do we need this dummy project? What happens if we just have `currentProject = null`?

client/src/views/Projects.vue
```diff
@@ -170,21 +176,42 @@ export default defineComponent({
...
+        </v-card-title>
         <div class="flex-container">
           <v-card
             v-if="user.is_superuser"
             class="flex-card"
             style="flex-grow: 4;"
           >
-            <v-subheader>Settings</v-subheader>
+            <v-subheader v-if="isGlobal">
+              WARNING: Global imports will modify multiple projects.
```
zachmullen, line 210:
```suggestion
              WARNING: Global imports will modify all projects referenced in the import file.
```

miqa/core/conversion/import_export_csvs.py
```diff
@@ -56,8 +62,11 @@ def validate_import_dict(import_dict, project: Project):
         import_schema.validate(import_dict)
         import_dict = validate_file_locations(import_dict, project)
     except SchemaError:
-        raise APIException(f'Invalid format of import file {project.import_path}')
-    if project.global_import_export:
+        import_path = (
+            GlobalSettings.load().import_path if project == 'global' else project.import_path
```
zachmullen, line 66:
```suggestion
            GlobalSettings.load().import_path if project is None else project.import_path
```

miqa/core/rest/global_settings.py
```diff
@@ -0,0 +1,74 @@
...
+
+class GlobalSettingsViewSet(ViewSet):
+    def get_object(self):
+        return GlobalSettings.load()
+
+    def get_permissions(self):
+        if self.request.method == 'GET':
+            return [IsAuthenticated()]
+        else:
+            return [IsSuperUser()]
```
zachmullen, line 31: Are these supposed to be instances of the permission classes, or the classes themselves?

miqa/core/rest/global_settings.py
```diff
@@ -0,0 +1,74 @@
...
+        methods=['GET', 'PUT'],
+    )
+    def settings_(self, request):
+        if request.method == 'PUT':
+            global_settings = self.get_object()
+            global_settings.import_path = request.data['importPath']
+            global_settings.export_path = request.data['exportPath']
+            global_settings.full_clean()
+            global_settings.save()
+        return Response(GlobalSettingsSerializer(self.get_object()).data)
```
zachmullen, line 54:
```suggestion
        global_settings = self.get_object()
        if request.method == 'PUT':
            global_settings.import_path = request.data['importPath']
            global_settings.export_path = request.data['exportPath']
            global_settings.full_clean()
            global_settings.save()
        return Response(GlobalSettingsSerializer(global_settings).data)
```

miqa/core/tasks.py
```diff
@@ -44,38 +45,39 @@ def evaluate_data(frame_ids, project_id):
         )
 
 
-def import_data(user_id, project_id):
-    project = Project.objects.get(id=project_id)
+def import_data(project_id: Optional[str]):
+    if not project_id:
```
zachmullen, line 49:
```suggestion
    if project_id is None:
```

miqa/core/tasks.py
```diff
@@ -44,38 +45,39 @@ def evaluate_data(frame_ids, project_id):
...
+        import_path = project.import_path
 
-    if project.import_path.endswith('.csv'):
-        import_dict = import_dataframe_to_dict(pandas.read_csv(project.import_path))
-    elif project.import_path.endswith('.json'):
-        import_dict = json.load(open(project.import_path))
+    if import_path.endswith('.csv'):
+        import_dict = import_dataframe_to_dict(pandas.read_csv(import_path))
+    elif import_path.endswith('.json'):
+        import_dict = json.load(open(import_path))
```
zachmullen, line 59: File handles should always be opened using a context manager, otherwise it's undefined behavior whether or not they'll be closed.

```suggestion
        with open(import_path) as import_fd:
            import_dict = json.load(import_fd)
```

miqa/core/tasks.py
```diff
@@ -44,38 +45,39 @@ def evaluate_data(frame_ids, project_id):
...
     new_scans = []
     new_frames = []
 
-    # comment out the below line and remove project_id param
-    # when multi-project imports are supported
-    project = Project.objects.get(id=project_id)
-
     for project_name, project_data in import_dict['projects'].items():
-        if project.global_import_export:
+        if not project_id:
```
zachmullen, line 75:
```suggestion
        if project_id is None:
```

miqa/core/tasks.py
```diff
@@ -106,41 +108,41 @@ def perform_import(import_dict, project_id):
     Scan.objects.bulk_create(new_scans)
     Frame.objects.bulk_create(new_frames)
 
-    evaluate_data.delay([frame.id for frame in new_frames], project.id)
+    frames_by_project = {}
+    for frame in new_frames:
+        project_id = str(frame.scan.experiment.project.id)
+        if project_id not in frames_by_project:
+            frames_by_project[project_id] = []
+        frames_by_project[project_id].append(str(frame.id))
```
zachmullen, line 116: It would be fine, and slightly preferable, to pass around `UUID` types rather than converting them to strings, but that's out of scope of this PR. We can do a follow-up one to refactor that at some point.

miqa/core/tasks.py
```diff
@@ -106,41 +108,41 @@ def perform_import(import_dict, project_id):
...
 @shared_task
-def perform_export(project_id, project_names):
-    project = Project.objects.get(id=project_id)
+def perform_export(project_id: Optional[str]):
     data = []
 
-    if project.global_import_export:
-        # A global export should export all projects listed in the import file
-        projects = [Project.objects.get(name=project_name) for project_name in project_names]
+    if not project_id:
```
zachmullen, line 137:
```suggestion
    if project_id is None:
```

zachmullen (changes requested):

client/src/store/index.ts
annehaley: The pseudo-project allows us to store the import and export paths, and it lets us avoid a major refactoring. It's possible to store them in a different place and refactor related files to use new checks. We check in a lot of places whether `currentProject` is null, for example, if `currentProject` is null we display "Select a Project" on the homepage. Should I make this change and refactor?

miqa/core/rest/global_settings.py
annehaley: It appears to work either way, but since the DRF docs use the classes themselves, I'll make the change.

annehaley: Changed in [93922ab](https://github.com/OpenImaging/miqa/pull/294/commits/93922abd8d5a2b039cbe50e012cc63e1b8f4ef8a)

client/src/store/index.ts
zachmullen: Would the refactor be making import and export paths into computed properties to abstract this away? Or is there more to it?

miqa/core/rest/global_settings.py
annehaley: Actually, scratch that. The permission classes should be uninstantiated when specified as `permission_classes=[cls1, cls2]` like in the docs, but when overriding the `get_permissions` function on a Viewset, the return value should be a list of permission class instances. The change in [93922ab](https://github.com/OpenImaging/miqa/pull/294/commits/93922abd8d5a2b039cbe50e012cc63e1b8f4ef8a) was a breaking change: [See tests](https://github.com/OpenImaging/miqa/runs/5055106927?check_suite_focus=true)

client/src/store/index.ts
annehaley: It just involves creating a new store variable and changing various `if` statements in the render. It's not too much so I made the change in [94b5f3a](https://github.com/OpenImaging/miqa/pull/294/commits/94b5f3a53811104e7720e22b1474cc59339bc2d3). Now we refer to global and project settings entirely separately.

miqa/core/tasks.py
annehaley: At least for now, these two new `str` casts can be removed without any trouble. Removed in [1b8c6fb](https://github.com/OpenImaging/miqa/pull/294/commits/1b8c6fb6cc9195bc005447d82623966ada82bc75)

zachmullen (approved):

PR #301, opened 2022-01-28
Documenting Import Methods

annehaley:
Resolves #301 

This is just a first pass at describing how to use our import system. Since documenting is  a delicate effort, I'd appreciate any tips for where I can elaborate, express something differently, reorganize information, or fix a typo.


README.md
```diff
@@ -96,3 +96,65 @@ Useful sub-commands include:
 
 To automatically reformat all code to comply with
 some (but not all) of the style checks, run `tox -e format`.
+
+
+## Importing / Exporting Data
```
dzenanz, line 101: This looks like documentation for how to use it. It is great! But I definitely agree with you that we should break up this huge main readme. We should separate it out into [development docs](https://github.com/OpenImaging/miqa/blob/master/development.md), [deployment docs](https://github.com/OpenImaging/miqa/blob/master/prod/README.md) and [usage docs](https://github.com/OpenImaging/miqa/blob/master/how-to-import-json.md). These files might need to be renamed or moved. https://github.com/OpenImaging/miqa/blob/master/prod/DEVELOPMENT.md might need to be merged with either development or deployment one? They should all be linked from the main readme, which should be quite short now, and be more an advertisement. I don't think we need to link to [NN docs](https://github.com/OpenImaging/miqa/blob/master/miqa/learning/README.md).

annehaley: Reorganized in 7f5e233

dzenanz (approved):
Looks good. I did not read all the documentation carefully.

prod/README.md
```diff
@@ -2,78 +2,115 @@
...
+
+2. Edit the newly-created `prod/.env` file and replace all instances of "miqa.local" with with the desired domain name of your MIQA instance.
+
+> You can also specify those variables explicitly in your local shell environment instead of editing the file. `docker-compose` will check for local shell values before defaulting to the values in `.env`.
+
+3. Edit your /etc/hosts file (or non-unix equivalent) and add the following entry (note that the separator is a tab). Replace `miqa.local` with the domain name of your MIQA instance.
+
+> 127.0.0.1 miqa.local
+
+4. You will need SSL certificates for your server. For a real production instance, these certificates should be signed by a third-parth certificate authority. If you have third-party certificates, skip to 5. For testing a production instance, it is sufficient to create self-sgined certificates.
```
zachmullen, line 24:
```suggestion
4. You will need SSL certificates for your server. For a real production instance, these certificates should be signed by a third-parth certificate authority. If you have third-party certificates, skip to 5. For testing a production instance, it is sufficient to create self-signed certificates.
```

prod/README.md
```diff
@@ -2,78 +2,115 @@
...
+## Create a superuser to administer the server
+**IMPORTANT**: Set both the Username and Email to the same valid email address. Admin logins will break if you do not.
 ```
 docker-compose run --rm django ./manage.py createsuperuser
 ```
 
-### Run the server
+
+## Run the server
+Run the following command in a new shell in the same directory (as this command needs to continue running as the server; use Ctrl+C to down the server).
```
zachmullen, line 63:
```suggestion
Run the following command in a new shell in the same directory (as this command needs to continue running as the server; use Ctrl+C to stop the server gracefully).
```

zachmullen (changes requested):

dchiquito (approved):
I added a few lines to specifically address the struggles Jim and Hans have been having. This is a great reorganization!

PR #302, opened 2022-02-02
Make Control Panel take up less vertical space

annehaley:


client/src/components/ControlPanel.vue
```diff
@@ -288,6 +298,29 @@ export default {
...
+                class="d-flex ml-5"
+                style="flex-direction:column"
+              >
+                <div style="flex-grow: 1">
+                  <v-switch
+                    :input-value="showCrosshairs"
+                    @change="setShowCrosshairs"
+                    label="Display crosshairs"
+                    hide-details
+                    class="shrink ma-0 pa-0 ml-n2"
```
dchiquito, line 311: These margings are small enough to impact usability:
image
The halo around the selected toggle is clickable, so once you have clicked a toggle, it overlaps the other toggle.

dchiquito (approved):
This is a much more usable layout on my screen.

It does feel a a little weird that comments are spread out across two columns. Something decision related has to be omitted from the rightmost column to keep the height homogenous, perhaps the autogenerated decisions would fit in the central column? Just a thought, I don't feel strongly.
image

client/src/components/ControlPanel.vue
annehaley: Changed in [4bd3d76](https://github.com/OpenImaging/miqa/pull/302/commits/4bd3d76a8a3e3b2464c53e951ffd12e659add88f)

annehaley:
> It does feel a a little weird that comments are spread out across two columns. Something decision related has to be omitted from the rightmost column to keep the height homogenous, perhaps the autogenerated decisions would fit in the central column? Just a thought, I don't feel strongly.

My original thought to this organization was separating read components from write components. If this change should be made, it will require a bit of refactoring to put various functions in the right places (currently `ControlPanel.vue` is mostly read functions and `DecisionButtons.vue` is mostly write functions.) It absolutely can be done,  but how strongly do you feel about this restructuring?

PR #304, opened 2022-02-03
Fix place crosshairs

annehaley:
During yesterday's demo, I realized that the version that fixed rotated images broke axis-swapped images. This fix works for both.


dchiquito (approved):
I confess most of this goes over my head, but the code looks good

dchiquito:
I just tried it out locally as well, everything seems to work perfectly with zoom, pan, and rotate :+1:

PR #312, opened 2022-02-07
Demo deployment support

zachmullen:


docker-compose.override.yml
```diff
@@ -7,7 +7,9 @@ services:
     command: ["./manage.py", "runserver", "0.0.0.0:8000"]
     # Log printing via Rich is enhanced by a TTY
     tty: true
-    env_file: ./dev/.env.docker-compose
+    env_file:
+      - ./dev/.env.docker-compose
+      - ./dev/.env.secrets
```
annehaley, line 12: Since `./dev/.env.secrets` is in the gitignore, instances without this file will not build. Is there a way this addition can be conditional? Or should we have a placeholder file?

docker-compose.override.yml
```diff
@@ -39,7 +41,9 @@ services:
     ]
     # Docker Compose does not set the TTY width, which causes Celery errors
     tty: false
-    env_file: ./dev/.env.docker-compose
+    env_file:
+      - ./dev/.env.docker-compose
+      - ./dev/.env.secrets
```
annehaley, line 46: Same as above

miqa/core/tasks.py
```diff
@@ -44,13 +47,30 @@ def evaluate_data(frame_ids, project_id):
...
     project = Project.objects.get(id=project_id)
 
     if project.import_path.endswith('.csv'):
-        import_dict = import_dataframe_to_dict(pandas.read_csv(project.import_path))
+        if project.import_path.startswith('s3://'):
+            buf = _download_from_s3(project.import_path).decode('utf-8')
+        else:
+            with open(project.import_path) as fd:
+                buf = fd.read()
+        import_dict = import_dataframe_to_dict(pandas.read_csv(buf))
```
annehaley, line 67: `read_csv` will read this as a file's name, not its contents. Instead you can pass in `StringIO(buf)` for pandas to read it as content.

annehaley (changes requested):
When attempting to load a Frame in the frontend using the `download` endpoint, the following server error occurs: 

```
django_1    | Traceback (most recent call last):
django_1    |   File "/usr/local/lib/python3.8/wsgiref/handlers.py", line 138, in run
django_1    |     self.finish_response()
django_1    |   File "/usr/local/lib/python3.8/wsgiref/handlers.py", line 183, in finish_response
django_1    |     for data in self.result:
django_1    |   File "/usr/local/lib/python3.8/wsgiref/util.py", line 37, in __next__
django_1    |     data = self.filelike.read(self.blksize)
django_1    | ValueError: read of closed file
``

File: miqa/core/models/frame.py [hunk 4]
```diff
@@ -38,3 +39,15 @@ def size(self) -> int:
...
+    @property
+    def is_in_s3(self) -> bool:
+        return self.raw_path.startswith('s3://')
+
+    @property
+    def s3_download_url(self) -> Optional[str]:
+        if not self.is_in_s3:
+            return None
+        bucket, key = self.raw_path.strip()[5:].split('/', maxsplit=1)
+        client = boto3.client('s3')
```
annehaley, line 52: Does this need auth at all?

zachmullen: The auth will come through the environment variables that will be set in the runtime environment

docker-compose.override.yml
zachmullen: Good point. For now I'll remove this until we find some robust way to handle this.

miqa/core/models/frame.py
```diff
@@ -38,3 +39,15 @@ def size(self) -> int:
...
+
+    @property
+    def is_in_s3(self) -> bool:
+        return self.raw_path.startswith('s3://')
+
+    @property
+    def s3_download_url(self) -> Optional[str]:
+        if not self.is_in_s3:
+            return None
+        bucket, key = self.raw_path.strip()[5:].split('/', maxsplit=1)
```
dchiquito, line 51: I generally use [urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) instead of using string functions. It is more verbose though so ¯\\\_(ツ)\_\/¯

dchiquito (commented):
I assume the facilities to get the data into S3 are coming later.

annehaley (approved):

PR #316, opened 2022-02-09
Allow anyone with project access to import/export

dchiquito:
Anyone with project access now is able to click the import/export buttons. They are not allowed to change import/export paths or the global import setting.

### DIFF
--- a/miqa/core/rest/project.py
+++ b/miqa/core/rest/project.py
@@ -172,7 +172,7 @@ class ProjectViewSet(
         request_body=no_body,
         responses={204: 'Import succeeded.'},
     )
-    @project_permission_required(superuser_access=True)
+    @project_permission_required()
     @action(detail=True, url_path='import', url_name='import', methods=['POST'])
     def import_(self, request, **kwargs):
         project: Project = self.get_object()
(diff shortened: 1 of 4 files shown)


annehaley (approved):
Looks good! Just needs a change to the test case now that imports are allowed by non-superusers.

miqa/core/rest/project.py
```diff
@@ -172,7 +172,7 @@ def settings_(self, request, **kwargs):
         request_body=no_body,
         responses={204: 'Import succeeded.'},
     )
-    @project_permission_required(superuser_access=True)
+    @project_permission_required()
```
annehaley, line 175: Not required, but you can pass in `review_access=True` if you think collaborators should not be given the same power of Import

dchiquito: It sounded like Killian wants everyone to be able to import, so I'm waiting on confirmation of that.

dchiquito: radio silence so far. I'd say we merge this as-is and we can lock it down later if appropriate.

zachmullen: In the interest of only opening it up as much as we need to, can we get away with just opening it to review_access? If so, that'd be my vote.

PR #318, opened 2022-02-10
Local uploads

### LINKED ISSUE #303: Allow remote upload to the server
Related to #179 and #87.

annehaley:
Resolves #303


miqa/core/tests/test_import.py
```diff
@@ -261,5 +261,5 @@ def test_import_s3_preserves_path(project_factory):
     frame = Frame.objects.first()
     assert (
         frame.raw_path
-        == 's3://miqa-sample/IXI_small/Guys/IXI002/0828-DTI/IXI002-Guys-0828-DTI-00.nii.gz'
+        == 's3:/miqa-sample/IXI_small/Guys/IXI002/0828-DTI/IXI002-Guys-0828-DTI-00.nii.gz'
```
zachmullen, line 264: We need the double slash to be preserved, see my comment above.

zachmullen (changes requested):

miqa/core/conversion/import_export_csvs.py
```diff
@@ -29,7 +29,7 @@ def validate_file_locations(input_dict, project):
                     raw_path = Path(import_path).parent.parent / raw_path
                 if not raw_path.exists():
                     raise APIException(f'Could not locate file "{raw_path}".')
-            input_dict[key] = value
+            input_dict[key] = str(raw_path)
```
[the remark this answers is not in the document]
annehaley: Ah ok, I was wondering why that change was made. Without this line, though, we don't get relative paths converted to absolute paths. For example, the IXI_small import csv has file paths relative to the location of the csv. Here, we use the location of the csv to convert the relative paths to absolute paths so the files can be found. 
Is there some way that we can satisfy both?

miqa/core/tests/test_import.py
annehaley: Ah ok, I thought it was a typo; I didn't understand the purpose of the test.

miqa/core/conversion/import_export_csvs.py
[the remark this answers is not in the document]
zachmullen: Ah, I missed that. First thing we should do, to prevent such oversights in the future, is to add a test for the relative path case, which should fail. Then, change the logic here to support both tests passing. It might involve an extra conditional somewhere, but that's OK as long as it satisfies all the behaviors we want to support.

If you notice any other behaviors in this method that don't have associated test cases, it'd be good to add them.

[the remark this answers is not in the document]
annehaley: Good idea. Another good reason for us to pay attention to code coverage soon

miqa/core/tests/test_import.py
annehaley: Changed in [1f06c2d](https://github.com/OpenImaging/miqa/pull/318/commits/1f06c2dfd89ee714109bc9566158e6088da4b9cd)

miqa/core/conversion/import_export_csvs.py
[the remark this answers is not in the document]
annehaley: Changed in [1f06c2d](https://github.com/OpenImaging/miqa/pull/318/commits/1f06c2dfd89ee714109bc9566158e6088da4b9cd)

miqa/core/tasks.py
```diff
@@ -130,12 +130,13 @@ def perform_import(import_dict, project_id: Optional[str]):
     Scan.objects.bulk_create(new_scans)
     Frame.objects.bulk_create(new_frames)
 
+    # must use str, not UUID, to get sent to celery task properly
```
zachmullen, line 133: Is this true? I would have assumed UUIDs would pickle up just fine

annehaley: Without the cast to string below that line, we get the following exception when we run an import: 
```
celery_1    | Traceback (most recent call last):
celery_1    |   File "/usr/local/lib/python3.8/site-packages/kombu/serialization.py", line 39, in _reraise_errors
celery_1    |     yield
celery_1    |   File "/usr/local/lib/python3.8/site-packages/kombu/serialization.py", line 210, in dumps
celery_1    |     payload = encoder(data)
celery_1    |   File "/usr/local/lib/python3.8/site-packages/kombu/utils/json.py", line 68, in dumps
celery_1    |     return _dumps(s, cls=cls or _default_encoder,
celery_1    |   File "/usr/local/lib/python3.8/json/__init__.py", line 234, in dumps
celery_1    |     return cls(
celery_1    |   File "/usr/local/lib/python3.8/json/encoder.py", line 199, in encode
celery_1    |     chunks = self.iterencode(o, _one_shot=True)
celery_1    |   File "/usr/local/lib/python3.8/json/encoder.py", line 257, in iterencode
celery_1    |     return _iterencode(o, 0)
celery_1    | TypeError: keys must be str, int, float, bool or None, not UUID
```

zachmullen: Huh, apparently JSON is kombu's default serializer. I think I'll bring this up on the platform channel to see if we want to use pickling instead.

zachmullen (approved):

PR #323, opened 2022-02-14
Fix timeout

dchiquito:
* Move `TimeoutDialog` so it always loads instead of only showing up on the Frame view.
* Adjust some timing and messaging for a cleaner UX.


annehaley (approved):
Works pretty well; I tried each of these three states once:
- on timeout, select "Continue Session"
- on timeout, select "Logout" and "Reload"
- on timeout, do nothing

PR #333, opened 2022-02-16
Add placeholder component when lacking edit access

### LINKED ISSUE #328: View-only mode placeholder for third column in Control Panel
When in view-only mode this column should contain:

1. An explanation of view-only mode and permissions
2. Who currently has the edit lock
2. A button to take the edit lock (only for superusers if another user has access, only for reviewers and superusers otherwise)

annehaley:
Resolves #328


miqa/core/rest/experiment.py
```diff
@@ -115,20 +115,24 @@ def lock(self, request, pk=None):
             experiment: Experiment = Experiment.objects.select_for_update().get(pk=pk)
             if experiment.project.archived:
                 raise ArchivedProject()
-            if experiment.lock_owner is not None and experiment.lock_owner != request.user:
+            if (
+                experiment.lock_owner is not None
+                and experiment.lock_owner != request.user
+                and not (request.user.is_superuser and 'force' in request.data)
```
dchiquito, line 121: This doesn't actually check the value of `force`, so if you included `force: False` it would count as being a forced lock.

dchiquito (approved):
Structurally LGTM, just a couple of quibbles.

miqa/core/rest/experiment.py
annehaley: Changed in [b069c7b](https://github.com/OpenImaging/miqa/pull/333/commits/b069c7b8a869904f0127648c41081bf4324b8c13)

PR #336, opened 2022-02-22
Fix window resize bug

### LINKED ISSUE #326: Changing window size breaks intersection lines
Resizing window after the app has bean loaded breaks the intersection lines and the ability to navigate by clicking.

Steps: load a scan. Go to next scan. Change window size. Crosshairs are wrong and click to navigate no longer works.

Tested on https://miqa.miqaweb.io/
Screenshot 2022-02-16 11 27 31

annehaley:
May resolve #326? I couldn't replicate the bug that the "next" button no longer works after resize.

### DIFF
--- a/client/src/components/VtkViewer.vue
+++ b/client/src/components/VtkViewer.vue
@@ -114,6 +114,7 @@ export default {
       view.getInteractor().onLeftButtonPress((event) => this.placeCrosshairs(event));
     },
     currentFrame() {
+      this.prepareViewer();
       this.representation.setSlice(this.slice);
     },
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
Code looks fine. I hope you tested it 😄

The "next" button worked, using it it just made it easier for me to reproduce the bug more reliably.

dzenanz:
Merge?

annehaley:
Yes, I tested all the variations of the bug that I could replicate. After merging, I'll test it on app.miqaweb.io where the bug was first seen.

PR #338, opened 2022-02-22
Fix Email component not calling send

annehaley:
Addresses #332


client/src/django.ts
```diff
@@ -190,7 +190,7 @@ const djangoClient = {
     const resp = await apiClient.get('/users');
     return resp.status === 200 ? resp.data : null;
   },
-  async sendEmail(email: string) {
+  async sendEmail(email: Object) {
```
zachmullen, line 193: While we're in here, could we make a proper interface type for this object?

PR #356, opened 2022-03-04
Refactor TimeoutDialog

### LINKED ISSUE #324: Refactor TimeoutDialog to avoid setTimeout
Currently, the `TimeoutDialog` is triggered using a `setTimeout`, and it then counts down from 2 minutes using 120 `setTimeout`s of 1 second each. This results in some janky behavior when the page isn't active (i.e. the user switched tabs, minimized the window, is using another application, etc) and the page isn't getting the cycles it needs to count the timeouts accurately. After waiting for 1 minute after the timeout should have triggered while switched to a different application, the timeout only counted down 30 seconds.

The server side timeout should still lock users out after 30 minutes, but the UI can get into a weird state where the server has timed out but the UI still thinks that there's a chance to refresh the login.

I think we should refactor the system so that the store keeps the last time a request was made, rather than using `setTimeout` to trigger a callback. `TimeoutDialog` will use a 1s `setTimeout` loop to check how long it's been since the last request. If 28 min < that time < 30 min, then the dialog is enabled and the text it shows is updated appropriately. If that time >= 30 minutes, then the server has already expired the session, so the client logs out immediately.
This way, even if the tab is completely inactive, whenever the user comes back the timeout will resolve and the UI will put itself into the correct login state.

dchiquito:
Fixes #324 

This is a complete rewrite of `TimeoutDialog` to use composition API and to forcibly log the user out much more robustly.

I used a `function foo() { setTimeout(foo, 1000) }` loop to check if timeouts have expired and to update the countdown. It doesn't feel very clean but I'm not sure how else to run something every second.


client/src/components/TimeoutDialog.vue
```diff
@@ -1,104 +1,105 @@
...
+        }
+      },
+    });
+    idle.start();
+
+    // This function calls itself after 1 second to check if the session has expired and to keep
+    // the countdown timer up to date.
+    const updateCountdown = () => {
+      const now = Date.now();
+      console.log('updating countdown', now);
```
annehaley, line 62: Remove?

annehaley: Remove console log?

client/src/components/TimeoutDialog.vue
```diff
@@ -1,104 +1,105 @@
...
+      timeRemainingStr,
+      reset,
+      logout,
+      sessionTimeout,
+      idleTimeout,
+    };
   },
-};
+  onIdle() {
+    // TODO why this does not trigger
```
annehaley: This seems like a blocker; what is this function meant to do and is it still not happening?

dchiquito: This was the idle-vue hook that I could not get to trigger. I opted to replace idle-vue with idle-js, which was more straightforward to set up and didn't involve any Vue magic.

annehaley (approved):

Issue #343, opened 2022-02-24
NN not running automatically in deployment


dzenanz:
There are no errors, but automatic evaluations are not present when reviewing the data. Happens on both SRI and Heroku instances.

annehaley:
@zachmullen now that #360 is resolved, could you perform an import on the prod instance (I don't have permission) and post any celery output so we can identify the issue?

annehaley:
We identified the issue that the `evaluate_data` task contains the following check:
```
if Path(frame.raw_path).exists():
```
This is not set up to support an s3 connection. @zachmullen suggests we download the file contents to a temporary directory before running the evaluation on them.

annehaley:
@zachmullen resolved the issue and now evaluations are available on https://miqa.miqaweb.io/#/

PR #363, opened 2022-03-08
Download s3 files before evaluation

annehaley:
Relates to #343. I'll merge and test on the prod deployment before closing that issue.

### DIFF
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -72,8 +72,7 @@ def evaluate_data(frames_by_project):
                     tmp = tempfile.NamedTemporaryFile(prefix=tmpdirname)
                     tmp.write(_download_from_s3(file_path))
                     file_path = tmp.name
-                else:
-                    file_paths.append(file_path)
+                file_paths.append(file_path)
             results = evaluate_many(current_model, file_paths)
 
             Evaluation.objects.bulk_create(
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -1,6 +1,7 @@
 from io import BytesIO, StringIO
 import json
 from pathlib import Path
+import tempfile
 from typing import Optional
 
 import boto3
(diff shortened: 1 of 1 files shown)


miqa/core/tasks.py
```diff
@@ -46,34 +55,37 @@ def evaluate_data(frames_by_project):
...
+        for model_name, frame_set in model_to_frames_map.items():
+            current_model = available_evaluation_models[model_name].load()
+            file_paths = []
+            for frame in frame_set:
+                file_path = str(frame.raw_path)
+                if file_path.startswith('s3://'):
+                    tmp = tempfile.NamedTemporaryFile(prefix=tmpdirname)
+                    tmp.write(_download_from_s3(file_path))
+                    file_path = tmp.name
+                else:
```
dzenanz: There should be no else. Otherwise we need something like `file_paths.append(tmp.name)` on line 74.

annehaley: Nice catch. Changed in [79415e6](https://github.com/OpenImaging/miqa/pull/363/commits/79415e6e8bd6f7d73c44cf05fc17152e5bbeb118)

miqa/core/tasks.py
```diff
@@ -72,8 +72,7 @@ def evaluate_data(frames_by_project):
                     tmp = tempfile.NamedTemporaryFile(prefix=tmpdirname)
                     tmp.write(_download_from_s3(file_path))
                     file_path = tmp.name
```
dzenanz: Do we need to update `results=results[str(frame.raw_path)]` on line 83?

annehaley: You're right. would something like [20e5673](https://github.com/OpenImaging/miqa/pull/363/commits/20e567360592c3a4f2e2af28946f4539fd3f95b5) work?

dzenanz (approved):
Preferably squash before or upon merging.

PR #355, opened 2022-03-03
Project creation

annehaley:
Resolves #354

PR #374, opened 2022-03-14
Display import buttons

annehaley:


zachmullen (approved):
Just so I understand -- the backend already supported anyone being able to import/export, it's just the front-end had not caught up?

annehaley:
> Just so I understand -- the backend already supported anyone being able to import/export, it's just the front-end had not caught up?

It was a regression with #355; adding this `v-if` for that PR was incorrect; I didn't account for Kilian's previous request that any user on the project should be able to import & export.

PR #378, opened 2022-03-15
Include User Guide in Github pages

### LINKED ISSUE #329: Develop user manual for MIQA

annehaley:
I'll need help getting these docs to render on a Github.io page like [DIVE does](https://kitware.github.io/dive/)

Resolves #329


dchiquito:
I'm pretty sure the DIVE docs are build with https://www.mkdocs.org/getting-started/, doesn't look too difficult to get set up. The DIVE docs are deployed with this bit of magic: https://github.com/Kitware/dive/blob/main/.github/workflows/release.yml#L72-L78
I think it really may be as simple as setting up MkDocs, pasting that snippet into our GitHub workflow that runs on merge to master and switching on GitHub pages. That should probably be done in a separate branch+PR though, since GitHub workflows never work the first time and make a big mess in the git history when debugging.

dchiquito (approved):

PR #379, opened 2022-03-16
Evaluate uploads

### LINKED ISSUE #353: Run evaluations async on locally uploaded scans in a project
* We could do poll/refresh to get the status of the update on the val

annehaley:
Resolves #353

### DIFF
--- a/client/src/components/DecisionButtons.vue
+++ b/client/src/components/DecisionButtons.vue
@@ -160,8 +160,10 @@ export default {
     ]),
     async pollForEvaluation() {
       const frameData = await djangoRest.frame(this.currentViewData.currentFrame.id);
-      this.setFrameEvaluation(frameData.frame_evaluation);
-      clearInterval(this.pollInterval);
+      if (frameData.frame_evaluation) {
+        this.setFrameEvaluation(frameData.frame_evaluation);
+        clearInterval(this.pollInterval);
+      }
     },
     convertValueToLabel(artifactName) {
       return artifactName
--- a/client/src/components/DecisionButtons.vue
+++ b/client/src/components/DecisionButtons.vue
@@ -145,10 +145,24 @@ export default {
       }
     });
   },
+  mounted() {
+    if (!this.currentViewData.currentAutoEvaluation) {
+      this.pollInterval = setInterval(this.pollForEvaluation, 1000 * 30);
+    }
+  },
+  beforeUnmount() {
+    clearInterval(this.pollInterval);
+  },
   methods: {
     ...mapMutations([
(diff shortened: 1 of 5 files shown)


client/src/components/DecisionButtons.vue
```diff
@@ -117,10 +117,24 @@ export default {
       this.warnDecision = false;
     },
   },
+  mounted() {
+    if (!this.currentViewData.currentAutoEvaluation) {
+      this.pollInterval = setInterval(this.pollForEvaluation, 1000 * 30);
```
dzenanz: I would rather have 10 seconds as the polling interval. 30 seems long.

dzenanz (approved):
Code looks good on a glance. I have not tried it.

dzenanz (approved):

client/src/components/DecisionButtons.vue
```diff
@@ -117,10 +117,24 @@ export default {
...
+  },
   methods: {
     ...mapMutations([
       'updateExperiment',
+      'setFrameEvaluation',
     ]),
+    async pollForEvaluation() {
+      const frameData = await djangoRest.frame(this.currentViewData.currentFrame.id);
+      this.setFrameEvaluation(frameData.frame_evaluation);
+      clearInterval(this.pollInterval);
```
dchiquito: This looks like it will stop the interval loop immediately after the first evaluation. Are you relying on an exception being thrown somewhere to stop that from happening? If so an explicit try/catch would be nice

miqa/core/rest/frame.py
```diff
@@ -81,7 +80,12 @@ class Meta:
         fields = ['content', 'scan']
 
 
-class FrameViewSet(ListModelMixin, GenericViewSet, mixins.CreateModelMixin):
+class FrameViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    GenericViewSet,
+    mixins.CreateModelMixin,
```
dchiquito, line 86: I think this goes before `GenericViewSet`, although I confess I'm not entirely sure what it would break.

client/src/components/DecisionButtons.vue
annehaley: The intention is that there's only ever one NN evaluation that we need to fetch, so we can stop polling once we get it.

annehaley: After getting clarification, I understand the point made here better and I added an if statement in [8fe070a](https://github.com/OpenImaging/miqa/pull/379/commits/8fe070a46eda4753de737552ccc07321713049e3) to address this

PR #389, opened 2022-03-17
Always redirect to login after timer expires

dchiquito:
The error that was causing the timeouts to keep counting down after reaching 0 was because the call to logout was failing because the user was already logged out. This will always redirect the user to the login page, regardless of whether or not they were already logged out.

### DIFF
--- a/client/src/components/TimeoutDialog.vue
+++ b/client/src/components/TimeoutDialog.vue
@@ -1,10 +1,10 @@
 <script lang="ts">
 import IdleJS from 'idle-js';
-import djangoRest from '@/django';
-import store from '@/store';
 import {
   computed, defineComponent, ref,
 } from '@vue/composition-api';
+import djangoRest from '@/django';
+import store from '@/store';
 
 const warningDuration = 2 * 60 * 1000; // the warning box will pop up for 2 minutes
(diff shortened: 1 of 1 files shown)


client/src/components/TimeoutDialog.vue
```diff
@@ -38,9 +38,13 @@ export default defineComponent({
...
     const logout = async () => {
-      await djangoRest.logout();
-      // This will redirect to the login page
-      await djangoRest.login();
+      try {
+        // This may fail with a 401 if the user has already been logged out
+        await djangoRest.logout();
+      } finally {
+        // This will redirect to the login page
+        await djangoRest.login();
```
zachmullen, line 46: If they're logged into the server, won't this just automatically re-log them into the client? If so, doesn't that defeat the purpose of the auto-logout?

dchiquito: `djangoRest.logout` invokes a special API endpoint that explicitly terminates the server session. If that call succeeds, then the user is no longer logged in when `djangoRest.login` is called, so they get sent to the login page. If that call fails, then the user was not logged in, so they will still get sent to the login page.

annehaley (approved):

PR #390, opened 2022-03-18
Interface adjustments

### LINKED ISSUE #375: Add Keyboard shortcut functionality back
During the control panel redesign, some keyboard shortcuts were made no longer applicable and were removed. We should add new shortcuts back that allow a user to navigate MIQA without a mouse. 

Below are the keyboard shortcuts that were deprecated. Some of these may need new equivalents.
```
['Increase/decrease window', [['=', '-']]],
['Increase/decrease window level', [['[', ']']]],
['Focus to note', [['n']]],
['Show note history', [['h']]],
['Unfocus from note', [['esc']]],
['Mark as bad/good/usable extra', [['b', 'g', 'u']]],
['Save', [['alt + s']]],
['Save on confirm dialog', [['y']]],
["Don't save on confirm dialog", [['n']]],
['Cancel on confirm dialog', [['esc']]],
```

### LINKED ISSUE #383: Allow a user to specify artifact presence instead of supplying a comment
@annehaley feel free to edit the title

### LINKED ISSUE #385: Import and export paths are not fully visible when they do not fit the space
Probably the best solution would be to have the more standard edit box control, which will slide the displayed string left if the user moves the cursor right.

We should also do import and export path in separate lines to significantly alleviate this problem.

### LINKED ISSUE #386: Enable xyz sliders to continuously update the view (current the views are updated on mouse up)

### LINKED ISSUE #387: Consistent error messages
Some errors have a modal error dialog, others have short-lived floating message. All errors should have a modal dialog.

annehaley:
This PR addresses several small interface issues all at once.
Resolves #375, #383, #385, #386, #387

### DIFF
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -56,11 +56,10 @@ export default defineComponent({
           await loadProject(currentProject.value);
         }
       } catch (ex) {
+        const text = ex || 'Import failed due to server error.';
+        importErrors.value = true;
+        importErrorText.value = text;
         importing.value = false;
-        this.$snackbar({
-          text: ex || 'Import failed. Refer to server logs for details.',
-        });
-        console.error(ex);
       }
       importDialog.value = false;
     }
@@ -76,12 +75,11 @@ export default defineComponent({
           text: 'Saved data to file successfully.',
           timeout: 6000,
         });
-      } catch (err) {
-        this.$snackbar({
-          text: err || `Export failed: ${err.response.data.detail || 'Server error'}`,
-          timeout: 6000,
-
-        });
+      } catch (ex) {
+        const text = ex || 'Export failed due to server error.';
+        importErrors.value = true;
+        importErrorText.value = text;
+        importing.value = false;
       }
       exporting.value = false;
     }
--- a/client/src/components/DecisionButtons.vue
+++ b/client/src/components/DecisionButtons.vue
@@ -120,6 +120,12 @@ export default {
     newComment() {
       this.warnDecision = false;
     },
+    confirmedPresent() {
+      this.warnDecision = false;
+    },
+    confirmedAbsent() {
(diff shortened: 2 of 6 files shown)


client/src/components/DecisionButtons.vue
```diff
@@ -120,6 +120,12 @@ export default {
     newComment() {
       this.warnDecision = false;
     },
+    confirmedPresent() {
+      this.warnDecision = false;
+    },
+    confirmedAbsent() {
+      this.warnDecision = false;
```
dzenanz, line 127: If all artifacts are cycled back to unconfirmed state, is that case handled?

dzenanz (approved):
Looks good on a glance. I have not tried it.

client/src/components/DataImportExport.vue
```diff
@@ -56,11 +56,10 @@ export default defineComponent({
...
         }
       } catch (ex) {
+        const text = ex || 'Import failed due to server error.';
+        importErrors.value = true;
+        importErrorText.value = text;
         importing.value = false;
-        this.$snackbar({
-          text: ex || 'Import failed. Refer to server logs for details.',
-        });
-        console.error(ex);
```
dchiquito, line 63: I'd keep this around, seems handy

dchiquito (approved):

client/src/components/DecisionButtons.vue
annehaley: In that case, the length of the `confirmedPresent` and `confirmedAbsent` lists would be 0, so the check at the top of `handleCommentSave` would catch and tell the user to make a comment.

client/src/components/DataImportExport.vue
annehaley: The same error message (`ex`) would be shown in the dialog instead of the temporary snackbar. Should we display both?

PR #392, opened 2022-03-23
Add Scan Decisions to export and import csvs

annehaley:


miqa/core/conversion/import_export_csvs.py
```diff
@@ -72,29 +85,38 @@ def validate_import_dict(import_dict, project: Optional[Project]):
 
 
 def import_dataframe_to_dict(df):
-    if list(df.columns) != IMPORT_CSV_COLUMNS:
+    if list(df.columns) != IMPORT_CSV_COLUMNS and list(df.columns) != IMPORT_CSV_COLUMNS[:6]:
```
zachmullen, line 91: This should at least have a code comment explaining that there are two valid CSV formats.

miqa/core/tasks.py
```diff
@@ -204,15 +240,28 @@ def perform_export(project_id: Optional[str]):
...
+            row_data = [
+                project_object.name,
+                frame_object.scan.experiment.name,
+                frame_object.scan.name,
+                frame_object.scan.scan_type,
+                frame_object.frame_number,
+                frame_object.raw_path,
+            ]
+            last_decision = frame_object.scan.decisions.last()
+            if last_decision:
```
zachmullen, line 269: A code comment explaining this encoding scheme would also be a good idea.

zachmullen (changes requested):

miqa/core/tasks.py
annehaley: Added in [0ed474f](https://github.com/OpenImaging/miqa/pull/392/commits/0ed474f3cad06e516fd43dc99574db1cc8a55d7f)

miqa/core/conversion/import_export_csvs.py
annehaley: Added in [0ed474f](https://github.com/OpenImaging/miqa/pull/392/commits/0ed474f3cad06e516fd43dc99574db1cc8a55d7f)

annehaley:
@zachmullen can you re-review (I added support to save crosshair location of decision to csv and fixed some bugs I found in testing decision saving)

zachmullen (approved):

PR #395, opened 2022-03-23
Add girder-pytest-pyppeteer

dchiquito:
* Install girder-pytest-pyppeteer using https://girder-pytest-pyppeteer.readthedocs.io/en/latest/setup/
* Add a test for setting project permissions using the project settings page
* Add a polyfill to the client to deal with pyppeteer's out of date chromium version

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -1,6 +1,7 @@
 import Vue from 'vue';
 import VueCompositionAPI from '@vue/composition-api';
 import Vuetify from 'vuetify';
+import 'polyfill-object.fromentries';
 
 import AsyncComputed from 'vue-async-computed';
 import config from 'itk/itkConfig';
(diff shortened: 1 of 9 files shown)


client/src/main.ts
```diff
@@ -1,6 +1,7 @@
 import Vue from 'vue';
 import VueCompositionAPI from '@vue/composition-api';
 import Vuetify from 'vuetify';
+import 'polyfill-object.fromentries';
```
annehaley, line 4: What does this do?

annehaley (approved):
Very cool. I'd like to learn how to write more of these tests

client/src/main.ts
dchiquito: >The pyppeteer chromium version is quite out of date and is missing
`Object.fromEntries`. We use this in a few places so it is easier to
simply add the polyfill than to factor it out.

PR #399, opened 2022-03-24
Suggested changes to default email recipients feature

annehaley:
The changes suggested during this morning's meeting were as follows:

- move the default recipients from 'cc' to 'to'
- move self recipient from 'bcc' to 'cc'
- add the project name to email subjects
- make the save button next to the default email recipients input the same as the save button for the import and export paths
- value in default email recipients input should refresh upon switching projects
- default email recipients input should be disabled for users that are neither a superuser nor the project creator

### DIFF
--- a/client/src/components/EmailDialog.vue
+++ b/client/src/components/EmailDialog.vue
@@ -81,22 +81,22 @@ export default defineComponent({
         return;
       }
       this.selectedScreenshots = [];
-      this.toCandidates = [];
-      this.ccCandidates = this.currentProject.settings.default_email_recipients.map(
+      this.toCandidates = this.currentProject.settings.default_email_recipients.map(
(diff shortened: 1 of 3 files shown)


zachmullen:
There's one other feature, which could go on a separate PR if you prefer, which is to set the `Reply-To` header in the email that is sent to be the email address of the originating user.

annehaley:
> There's one other feature, which could go on a separate PR if you prefer, which is to set the `Reply-To` header in the email that is sent to be the email address of the originating user.

Thanks for reminding me, I added that in [886c626](https://github.com/OpenImaging/miqa/pull/399/commits/886c626c0af07b3a61f6b9824fa661e5a96b7ca8)

zachmullen (approved):

PR #411, opened 2022-03-31
Enable Codecov via CI

zachmullen:


zachmullen:
I'll go ahead and merge this. We can tweak the info included in the comment by adding codecov.yml if we want, but this is a reasonable first pass.

The reason for the decrease here is that my manual upload didn't omit `tests/` and `migrations/` directories.

PR #412, opened 2022-03-31
New release fixes

### LINKED ISSUE #402: Duplicate scan produces no error
Trying to upload of a scan via "Add scans"  into a new experiment with a name which already exists does not display an error. Instead "nothing happens". Clicking "upload" produces no visible action. It does not matter whether the file(s) in question already exist.
Screenshot 2022-03-30 09 39 07

### LINKED ISSUE #404: Upload scans improvements
* Show error message to the user, currently it is silent 
* Cancel button on the dialog

annehaley:
Resolves #404, resolves #402
This also addresses some errors raised in sentry for miqaweb and addresses a concern raised by Killian that needing to select screenshots to include as email attachments is unintuitive.

### DIFF
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -157,6 +157,13 @@ export default defineComponent({
       width="500"
     >
       <v-card>
+        <v-btn
+          icon
+          style="float:right"
+          @click="importDialog=false"
+        >
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
         <v-card-title class="title">
           Import
         </v-card-title>
@@ -195,6 +202,13 @@ export default defineComponent({
         class="pa-5"
         style="overflow: auto"
       >
+        <v-spacer />
+        <v-btn
+          icon
+          @click="importErrors=false"
+        >
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
         <v-card-title class="title">
           Import Errors Encountered
         </v-card-title>
--- a/miqa/core/rest/exceptions.py
+++ b/miqa/core/rest/exceptions.py
@@ -17,7 +17,7 @@ def custom_exception_handler(exc, context):
     # If an excpetion is unexpected, it will not be an APIException
     if not isinstance(exc, APIException):
         exception_identifier = uuid.uuid4()
-        logger.exception(f'Error {exception_identifier}: {exc} / {context}')
+        logger.exception(f'Unexpected REST API error: {exception_identifier}')
         return Response(
             data={
                 'detail': 'Unexpected server error. '
--- a/miqa/core/rest/exceptions.py
+++ b/miqa/core/rest/exceptions.py
@@ -1,6 +1,7 @@
 import logging
 import uuid
 
+from django.http.response import Http404
 from rest_framework import status
 from rest_framework.exceptions import APIException
 from rest_framework.response import Response
@@ -14,8 +15,7 @@ def custom_exception_handler(exc, context):
     # to get the standard error response.
     response = exception_handler(exc, context)
 
-    # If an excpetion is unexpected, it will not be an APIException
-    if not isinstance(exc, APIException):
+    if not isinstance(exc, APIException) and not isinstance(exc, Http404):
         exception_identifier = uuid.uuid4()
         logger.exception(f'Error {exception_identifier}: {exc} / {context}')
         return Response(
--- a/miqa/core/rest/exceptions.py
+++ b/miqa/core/rest/exceptions.py
@@ -1,0 +1,28 @@
+import logging
+import uuid
+from rest_framework import status
+from rest_framework.exceptions import APIException
+from rest_framework.views import exception_handler
+from rest_framework.response import Response
+
+
+logger = logging.getLogger('MIQA')
+
+
+def custom_exception_handler(exc, context):
+    # Call REST framework's default exception handler first,
+    # to get the standard error response.
+    response = exception_handler(exc, context)
+
+    # If an excpetion is unexpected, it will not be an APIException
+    if not isinstance(exc, APIException):
+        exception_identifier = uuid.uuid4()
+        logger.exception(f'Error {exception_identifier}: {exc} / {context}')
+        return Response(
+            data={
+                'detail': 'Unexpected server error. '
+                f'Refer to error {exception_identifier} in the server logs.'
+            },
+            status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        )
+    return response
--- a/miqa/core/rest/exceptions.py
+++ b/miqa/core/rest/exceptions.py
@@ -1,9 +1,7 @@
 import logging
 import uuid
 
-from django.http.response import Http404
 from rest_framework import status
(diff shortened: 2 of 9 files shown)


dzenanz (commented):
LGTM

miqa/core/rest/experiment.py
```diff
@@ -79,12 +80,18 @@ def create(self, request, *args, **kwargs):
...
+            experiment = Experiment(
+                project=project,
+                name=serializer.data['name'],
+                lock_owner=None,
+            )
+            experiment.save()
+        except utils.IntegrityError:
+            raise APIException('An experiment with the same name already exists in this project.')
+        except Exception as e:
+            print(type(e))
```
zachmullen: Should this block go away? Seems like we shouldn't be hiding unexpected exceptions here.

annehaley: We still raise an APIException with the exact text from the original exception. This is the catch-all that will raise any other types of errors to the user interface. The IntegrityError block just applies a custom error text instead of the default text about constraint violation. Although, I hadn't meant to leave the print statement in there...

zachmullen: I think that's not a good practice. If these errors are truly unexpected, we don't want to be sending the exception data to the end-user, because it could contain things we don't want to leak out. (Not to mention it's likely overly technical, and our developers who can make sense of that will be getting the technical details anyway from Sentry.)

I think we should do our best to try and capture any *known* invalid behaviors and wrap them up with a nice error message, but for truly unexpected exceptions, we need to just log them in Sentry and send a generic message to the user IMO.

miqa/core/rest/experiment.py
```diff
@@ -79,12 +80,18 @@ def create(self, request, *args, **kwargs):
...
-        experiment.save()
+        try:
+            experiment = Experiment(
+                project=project,
+                name=serializer.data['name'],
+                lock_owner=None,
+            )
+            experiment.save()
+        except utils.IntegrityError:
+            raise APIException('An experiment with the same name already exists in this project.')
```
zachmullen: I think we should call `full_clean` on the model and let the first-class validation logic form its own errors. This logic isn't very future proof; imagine if we add other check constraints or database validations to the model, then we don't know that an IntegrityError implies a name collision.

zachmullen (changes requested):

miqa/core/rest/experiment.py
annehaley: Okay, that's a good explanation. For now, we may have to tell Killian the same thing, that whatever exception he was getting is still unknown and he may need to tell us how to replicate the issue (if the problem was truly different than the experiment name constraint)

miqa/core/rest/experiment.py
zachmullen: Given that we are already calling `is_valid` above, we could add this validation logic into the serializer itself.

annehaley: Ok, I like that better. I'll make that change

miqa/core/rest/experiment.py
jeffbaumes: Could we output to the user, and log, a reference id (just a random number/string)? That way, Killian can tell us the reference number and we (with Jim's help) can look up the details of the error in the logs.

miqa/core/rest/experiment.py
annehaley: Changed in [2aee6e4](https://github.com/OpenImaging/miqa/pull/412/commits/2aee6e45b75d78842ce9b9543e2a3f8c486330e1)

miqa/core/rest/experiment.py
zachmullen: I like this idea and have done this on other projects by installing an exception handler generally across the API that handles otherwise uncaught `Exception`s. It creates a `uuid4` and sends it with the error message, and then calls `logger.exception()` including that UUID.

annehaley: Added in [11dd78c](https://github.com/OpenImaging/miqa/pull/412/commits/11dd78c7383708f92054ed5c4b1cf9b60288da83)

miqa/core/rest/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+def custom_exception_handler(exc, context):
+    # Call REST framework's default exception handler first,
+    # to get the standard error response.
+    response = exception_handler(exc, context)
+
+    # If an excpetion is unexpected, it will not be an APIException
+    if not isinstance(exc, APIException):
+        exception_identifier = uuid.uuid4()
+        logger.exception(f'Error {exception_identifier}: {exc} / {context}')
```
zachmullen:
```suggestion
        logger.exception(f'Unexpected REST API error: {exception_identifier}')
```

We should get the exception data for free by virtue of using `logger.exception`, and `context` should also be dumped to Sentry for free.

miqa/core/rest/exceptions.py
```diff
@@ -0,0 +1,28 @@
+import logging
+import uuid
+
+from rest_framework import status
+from rest_framework.exceptions import APIException
+from rest_framework.response import Response
+from rest_framework.views import exception_handler
+
+logger = logging.getLogger('MIQA')
```
zachmullen:
```suggestion
logger = logging.getLogger(__name__)
```

miqa/core/rest/experiment.py
```diff
@@ -49,6 +50,11 @@ class Meta:
         model = Experiment
         fields = ['name', 'project']
 
+    def validate(self, data):
+        if Experiment.objects.filter(name=data['name'], project=data['project']).count() > 0:
```
zachmullen:
```suggestion
        if Experiment.objects.filter(name=data['name'], project=data['project']).exists():
```

zachmullen (changes requested):

client/src/components/DataImportExport.vue
```diff
@@ -157,6 +157,13 @@ export default defineComponent({
       width="500"
     >
       <v-card>
+        <v-btn
+          icon
+          style="float:right"
```
zachmullen, line 162: I think using float:right is deprecated at this point, but I'm not enough of a vuetify or CSS expert to say exactly what to do instead. @dchiquito can you comment on this?

miqa/core/rest/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def custom_exception_handler(exc, context):
+    # Call REST framework's default exception handler first,
+    # to get the standard error response.
+    response = exception_handler(exc, context)
+
+    if not isinstance(exc, APIException) and not isinstance(exc, Http404):
```
zachmullen: The addition of the `Http404` check here made me a bit nervous, so I looked at the base `exception_handler`, and this behavior is documented:

> Any unhandled exceptions may return `None`, which will cause a 500 error to be raised.

So all we need to do here is:

```suggestion
    if response is None:
```

zachmullen (changes requested):

dchiquito (approved):

zachmullen (approved):

PR #416, opened 2022-04-04
Pyppeteer email test

dchiquito:
Add a new pyppeteer test for email functionality, specifically modifying recipients and attaching screenshots.

* Move the CSV/JSON file generation functions to a new `helpers.py`. IMHO fixtures are generally data, so we need a separate place for test helper/utility functions.
* Add a new fixture `samples_project`. It's a project that already has the samples imported.
* Add `PyppeteerTestingConfiguration`. It's the same as `DevelopmentConfiguration`, but with `CELERY_TASK_ALWAYS_EAGER=True`. See https://girder-pytest-pyppeteer.readthedocs.io/en/latest/debugging/#why-arent-my-celery-tasks-running


miqa/settings.py
```diff
@@ -78,6 +78,11 @@ class TestingConfiguration(MiqaMixin, TestingBaseConfiguration):
     CELERY_TASK_ALWAYS_EAGER = True
 
 
+class PyppeteerTestingConfiguration(MiqaMixin, DevelopmentBaseConfiguration):
+    # We would like to test that the celery tasks work correctly when triggered from the API
+    CELERY_TASK_ALWAYS_EAGER = True
```
zachmullen, line 92: Does the regular `TestingConfiguration` not work for this?

dchiquito: Excerpt from https://girder-pytest-pyppeteer.readthedocs.io/en/latest/setup/#add-the-test-pyppeteer-tox-environment:

>DJANGO_CONFIGURATION = DevelopmentConfiguration - This is a shortcut that uses a default [composed configuration](https://github.com/girder/django-composed-configuration) to conveniently set everything required for a more-or-less functioning server environment suitable for testing against. The TestingConfiguration omits several settings that are needed when running a realistic server, so the DevelopmentConfiguration is used instead. You may have an existing configuration that is more suitable, or you may want to set up your own custom PyppeteerTestingConfiguration for explicitly configuring the test environment.

zachmullen:
This is really cool.

Let's also add `{posargs:--cov=miqa --cov-report=xml}` to the pyppeteer tox command so we get code coverage, then this LGTM.

miqa/core/tests/pyppeteer/test_email.py
```diff
@@ -0,0 +1,123 @@
...
+        await page.waitForXPath(
+            f'//div[contains(@class,"v-card__text")]/span[.="{experiment.name}/{scan.name}/0/Sagittal"]'  # noqa: E501
+        )
+    ).click()
+    # Attach the third screenshot to the email
+    await (
+        await page.waitForXPath(
+            f'//div[contains(@class,"v-card__text")]/span[.="{experiment.name}/{scan.name}/0/Coronal"]'  # noqa: E501
+        )
+    ).click()
```
annehaley: I just merged #412 which makes every screenshot selected for attaching by default, which means this portion is likely unnecessary now

annehaley (approved):
Very cool, I'm already learning a lot about how to use pyppeteer

PR #422, opened 2022-04-06
Account creation

### LINKED ISSUE #420: User signup sheet should include fields for first name and last name
Right now this information is only editable in the admin console, which means the avatars for new users will not show their initials (everyone gets the default user icon avatar).

### LINKED ISSUE #421: Account activation process broken
When a new user joins and verifies their email address, a separate email is sent to all superusers saying this account should be set to active. Those emails sent to superusers include a link that doesn't work.

annehaley:
Resolves #421, resolves #420

### DIFF
--- a/miqa/core/rest/accounts.py
+++ b/miqa/core/rest/accounts.py
@@ -1,3 +1,5 @@
+from allauth.account.forms import SignupForm
+from django import forms
 from django.conf import settings
 from django.contrib.auth import logout
 from django.contrib.auth.models import User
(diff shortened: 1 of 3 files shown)


miqa/core/rest/accounts.py
```diff
@@ -82,3 +86,18 @@ def post(self, request):
...
+        self.fields['first_name'] = forms.CharField(
+            label=('First Name'),
+            min_length=1,
+            widget=forms.TextInput(attrs={'placeholder': ('First name')}),
+        )
+        self.fields['last_name'] = forms.CharField(
+            label=('Last Name'),
+            min_length=1,
+            widget=forms.TextInput(attrs={'placeholder': ('Last name')}),
+        )
```
dchiquito:
```suggestion
    first_name = forms.CharField(
        label=('First Name'),
        min_length=1,
        widget=forms.TextInput(attrs={'placeholder': ('First name')}),
    )
    last_name = forms.CharField(
        label=('Last Name'),
        min_length=1,
        widget=forms.TextInput(attrs={'placeholder': ('Last name')}),
    )
```
I think the only reason to assign fields in `__init__` is if they are conditional or if you need to process arguments somehow.

dchiquito (approved):

miqa/core/rest/accounts.py
annehaley: Cool, changed in [b16d311](https://github.com/OpenImaging/miqa/pull/422/commits/b16d311d9c04f339896d94f7087b55869e87c201)

PR #423, opened 2022-04-07
Omit uploaded scans from export csv and add explanatory dialog message

### LINKED ISSUE #415: Importing & exporting uploaded Scans
Scans which are uploaded (not imported) have remote locations (in minio or s3). When a Project containing Scans like these is exported, these Scans have a blank value for the `file_location`. 

We will need to add cyclical import/export support for Scans like these. If we do record the presigned s3 url as the `file_location` for Scans like these, how should they be imported? Should we download the file and reference its file path in the newly created Frame's `raw_path`? Or reference the same link in the newly created Frame's S3FileField `content`?

annehaley:
Resolves #415

### DIFF
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -66,10 +66,16 @@ export default defineComponent({
     async function exportData() {
       exporting.value = true;
       try {
+        let response;
         if (isGlobal.value) {
-          await djangoRest.globalExport();
+          response = await djangoRest.globalExport();
         } else {
-          await djangoRest.projectExport(currentProject.value.id);
+          response = await djangoRest.projectExport(currentProject.value.id);
+        }
+        if (response.warnings) {
+          importErrors.value = true;
+          importErrorText.value = response.detail;
+          importErrorList.value = response.warnings;
         }
         this.$snackbar({
           text: 'Saved data to file successfully.',
@@ -210,9 +216,11 @@ export default defineComponent({
           <v-icon>mdi-close</v-icon>
         </v-btn>
         <v-card-title class="title">
-          Import Errors Encountered
+          Messages Encountered
         </v-card-title>
-        {{ importErrorText }}
+        <div class="px-5">
+          {{ importErrorText }}
+        </div>
         <v-divider class="my-3" />
 
         <v-card-text
--- a/client/src/components/DataImportExport.vue
+++ b/client/src/components/DataImportExport.vue
@@ -72,7 +72,7 @@ export default defineComponent({
         } else {
           response = await djangoRest.projectExport(currentProject.value.id);
         }
-        if (response.warnings) {
+        if (!response || response.warnings) {
           importErrors.value = true;
           importErrorText.value = response.detail;
           importErrorList.value = response.warnings;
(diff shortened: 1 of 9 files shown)


client/src/django.ts
```diff
@@ -71,10 +71,10 @@ const djangoClient = {
     return (await apiClient.post(`/projects/${projectId}/import`)).data;
   },
   async globalExport() {
-    return apiClient.post('/global/export');
+    return (await apiClient.post('/global/export')).data;
   },
   async projectExport(projectId: string) {
```
dchiquito: It would be nice to have these methods typed so that it's obvious that the response is `{"detail":..., "warnings":...}` or `null`

client/src/components/DataImportExport.vue
```diff
@@ -66,10 +66,16 @@ export default defineComponent({
...
       try {
+        let response;
         if (isGlobal.value) {
-          await djangoRest.globalExport();
+          response = await djangoRest.globalExport();
         } else {
-          await djangoRest.projectExport(currentProject.value.id);
+          response = await djangoRest.projectExport(currentProject.value.id);
+        }
+        if (response.warnings) {
```
dchiquito: Is response `null` here, or `{}`? I would have expected `null` but idk

dchiquito (approved):

client/src/components/DataImportExport.vue
annehaley: The backend sends a `Response` object with `detail` and `warnings`. I don't think there's a way to get a null response from this endpoint.

client/src/django.ts
annehaley: Good idea. I added a new type `ResponseData` that we can use as the return type for functions in the django module which don't return some other type (POST endpoints, etc): [d487532](https://github.com/OpenImaging/miqa/pull/423/commits/d4875326cff2dcbfe4431b1b5d9b60bdb3643655)

client/src/components/DataImportExport.vue
dchiquito: It looks like it sends an empty 204 if there are no warnings. This `if` seems to be checking if there are any errors to display, implying that there is some sort of empty response indicating no errors.

client/src/types.ts
```diff
@@ -9,6 +9,12 @@ interface User {
   is_superuser: boolean
 }
 
+interface ResponseData {
+  detail: string,
+  errors: string[],
```
dchiquito, line 14: I think this would be optional

client/src/components/DataImportExport.vue
annehaley: In that case, `response` here is an empty string because it is returned from `projectExport` as `(await ...).data`. The if statement doesn't result in any errors, it just evaluates to false. But for the sake of clarity, I changed the if statement in [96aac5f](https://github.com/OpenImaging/miqa/pull/423/commits/96aac5f99fa1b7442159c1e9bd7bf11ad3d3afbc)

client/src/types.ts
annehaley: They are already optional in this interface definition. An object with only keys `detail, errors` is equally valid.

client/src/components/DataImportExport.vue
dchiquito: :+1: That should probably be included in the type definition

PR #428, opened 2022-04-12
Allow direct API authorization

annehaley:
When going over the admin manual, @curtislisle brought up an excellent question about MIQA API access. Since we don't have a dedicated Python client, we should expose the proper `django-oauth` endpoints so the user can fetch their own token and use that for subsequent API calls.

### DIFF
--- a/miqa/core/tests/test_rest.py
+++ b/miqa/core/tests/test_rest.py
@@ -17,12 +17,12 @@ def test_api_token_access(user, api_client):
     resp_1 = api_client.post(
         '/api-token-auth/',
         data={
-            "username": user.username,
-            "password": password,
+            'username': user.username,
+            'password': password,
         },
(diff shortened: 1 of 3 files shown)


zachmullen:
Also worth mentioning, if we do make a python client, we can use this approach: https://github.com/girder/girder-cli-oauth-client

dchiquito (changes requested):
For completeness, can you add a test that grabs a token using the new endpoint and uses it to access a permission-required endpoint?

annehaley:
Good idea @dchiquito , added in [d271a22](https://github.com/OpenImaging/miqa/pull/428/commits/d271a225405adfd8af3a8c1d53cc3f66a513afe2)

dchiquito (approved):

PR #440, opened 2022-04-14
Add link to documentation in navbar

annehaley:


client/src/components/Navbar.vue
```diff
@@ -120,6 +124,14 @@ export default defineComponent({
       />
     </div>
 
+    <v-btn
+      elevation="0"
```
dchiquito, line 128:
```suggestion
      elevation="0"
      class="mx-2"
```
Currently the button is touching the user icon, which is visible on hover:
image
A little padding would be nice

dchiquito (approved):

PR #446, opened 2022-04-18
Add experiment_notes column to import/export file formats

annehaley:


miqa/core/tasks.py
```diff
@@ -327,27 +333,6 @@ def perform_export(project_id: Optional[str]):
...
-                        )
-                    artifacts = last_decision.user_identified_artifacts.items()
-                    row_data += [
-                        last_decision.decision,
-                        last_decision.creator.email,
-                        last_decision.note.replace(',', ';'),
-                        last_decision.created,
-                        ';'.join([artifact for artifact, value in artifacts if value == 1]),
-                        location,
-                    ]
```
dchiquito, line 350: Was this just accidentally duplicated?

dchiquito (commented):
I don't see any changes to the import/export test. Is this tested implicitly somehow?

miqa/core/tasks.py
annehaley: yes, I think as the result of a merge.

annehaley:
> I don't see any changes to the import/export test. Is this tested implicitly somehow?

The experiment_notes column is optional, so it didn't affect the test we have, but I can add another test that uses all the optional columns.

EDIT: The new test revealed errors that I fixed, so it was helpful! :)

PR #449, opened 2022-04-20
Demo mode

### LINKED ISSUE #439: Onboarding workflow
Add "Try it" button to the landing page which goes to miqademo.

Add a "Documentation" button to the landing page that points to our overall documentation.

Add a "Documentation" button to the MIQA app header as well.

Add a demo account with demo data loaded. A script to reset that runs nightly would be great but could be a follow-on PR. It seems like this could simply be a project CSV which is imported each night.

Add a note to the login page instructing people how to log in with the demo account.

annehaley:
May resolve #439 
This PR creates a demo mode that we can use on `miqaweb`. In order to reset the state of the server, we only need to run `python manage.py populate_demo` in the django container on the server. Right now, we can only do this manually, but we can look into automation options on Heroku later.


miqa/core/management/commands/populate_demo.py
```diff
@@ -0,0 +1,39 @@
...
+import djclick as click
+from guardian.shortcuts import assign_perm
+
+from miqa.core.models import Project
+from miqa.core.tasks import import_data
+
+
+@click.command()
+def command():
+    # Delete extra projects in the system that users may have created
```
dchiquito: This should check the value of `DEMO_MODE` before doing anything else

miqa/core/management/commands/populate_demo.py
```diff
@@ -0,0 +1,39 @@
...
+from django.contrib.auth.models import User
+import djclick as click
+from guardian.shortcuts import assign_perm
+
+from miqa.core.models import Project
+from miqa.core.tasks import import_data
+
+
+@click.command()
+def command():
```
dchiquito, line 7: I think it's generally a good practice to treat CLI commands like this as an API surface and restrict the logic in them to handling CLI arguments and logging, then move the actual business logic stuff of resetting the server to a more centrally useful place. It keeps the CLI details out of the actual impactful code, and makes it easier later to do the reset from a scheduled celery task or from an API call or whatever we might want to do in the future.

In this case, this function would literally just call a different function in a different file and nothing else, so there's not a lot of value in doing that now.

miqa/core/rest/accounts.py
```diff
@@ -88,6 +89,22 @@ def post(self, request):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
+class DemoModeLoginView(LoginView):
```
dchiquito, line 92: It would be a better UX to modify the login template directly to include a new `<form>` with a button that will log the demo user in directly without having to enter any credentials.

miqa/core/tasks.py
```diff
@@ -137,7 +137,9 @@ def import_data(project_id: Optional[str]):
             else:
                 with open(import_path) as fd:
                     buf = fd.read()
-            import_dict = import_dataframe_to_dict(pandas.read_csv(StringIO(buf)))
+            import_dict = import_dataframe_to_dict(
+                pandas.read_csv(StringIO(buf)).fillna('').astype('string')
```
dchiquito, line 141: I've been using `pandas.read_csv(StringIO(bug), na_filter=False)`. I'm not entirely sure what these flags do though, so I'm not sure if that is equivalent or not.

dchiquito (changes requested):

miqa/core/management/commands/populate_demo.py
annehaley: Changed in [5cb08e8](https://github.com/OpenImaging/miqa/pull/449/commits/5cb08e840ddb8115d0c096db10c908b61b2202ea)

miqa/core/management/commands/populate_demo.py
annehaley: Changed in [5cb08e8](https://github.com/OpenImaging/miqa/pull/449/commits/5cb08e840ddb8115d0c096db10c908b61b2202ea)

jeffbaumes (approved):
@annehaley and I went over this and I suggested some minor changes to the login screen and reset behavior (only reset/delete projects belonging to the demo account). Once those are in and @dchiquito approves this is good to go.

hendu25:
FYI, I've just gotten access to the demo system but don't have access to view anything.  Instead of filling a new issue, just commenting here in case this PR fixes this problem

zachmullen (commented):
Go ahead and ship this if others are OK with it. I'll take a closer look next week, nothing in here seems too risky to merge at this point.

miqa/core/rest/accounts.py
annehaley: Added in [83c4e79](https://github.com/OpenImaging/miqa/pull/449/commits/83c4e79006c654ab184c070e51f168e12e3b54ff)

Issue #450, opened 2022-04-20
Pyppeteer test is non-deterministic


annehaley:
I've gotten a failure like this a few times on some PRS: https://github.com/OpenImaging/miqa/runs/6101270717?check_suite_focus=true

After re-running the job, it passes. Perhaps some race condition is created and the expected outcome only happens some percentage of the time? Specifically, the failure only happens at the point where the test ensures that the default email recipient is present in the `To` field.

PR #468, opened 2022-04-26
Cache lfs, pip deps, and node deps for CI

annehaley:


.github/workflows/cache.yml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  cache-lfs:
+    runs-on: ubuntu-latest
+
+    - name: Cache git lfs
+      uses: actions/cache@v3
```
dzenanz: My first impression was that this was supposed to be added to existing `.yml` files. Will a separate `cache.yml` work?

annehaley: No, you're right. This first attempt was not successful. I'm looking into sharing one built image between jobs right now.

annehaley:
I've been looking around for a way to test this locally, but I haven't found anything useful. I did some significant refactoring and the tests don't run now, so I need a way to diagnose.

dzenanz:
The usual way is to have the local setup close to what is being done inside CI virtual machine.

annehaley:
I found `act`, which is a good way to test Github Actions locally: https://github.com/nektos/act. Using this tool, I was able to run this multiple times and find the most efficient configuration. 

Additionally, I reconfigured the job to not require a successful download from git-lfs. It will try, but if that failure happens the tests will continue. The celery task will still spawn in the pytest (we don't currently test what happens within the celery task) and it will raise the `UnpicklingError` we know and love. We can write a test later that can provide insights about the success of the celery task (without making the whole run crash like we see here https://github.com/OpenImaging/miqa/runs/6165918974?check_suite_focus=true)

The only failure we're getting now is known: #450

.github/workflows/ci.yml
```diff
@@ -22,8 +33,6 @@ jobs:
         ports:
           - 5672:5672
       minio:
-        # This image does not require any command arguments (which GitHub Actions don't support)
-        # TODO at the time this was pinned, latest bitnami/minio was broken
```
dchiquito: Why remove this? It still seems pertinent to me, unless you meant to change the image to `bitnami/minio:latest`.

.github/actions/cache_deps/action.yml
```diff
@@ -0,0 +1,59 @@
+name: Cache dependencies
+inputs:
+  fetch-lfs:
+    default: false
+  download-node-deps:
+    default: false
```
dchiquito: Why not have three actions, `lfs.yml`, `tox.yml`, and `npm.yml`? If a CI step needs multiple it can just invoke the ones it needs, and you don't need to worry about these toggle arguments.

dzenanz (commented):
I am not familiar enough with GitHub actions to properly judge this.

.github/workflows/ci.yml
annehaley: Oh you're right. In a previous commit, I bumped it to latest and removed the comment, but then in my last commit, I pinned it again because the latest minio still breaks our `check-migrations` test. It at least was a good checkpoint for whether we still need it pinned (we do). So I'll put the comment back in, too.

annehaley: Changed in [fd836a4](https://github.com/OpenImaging/miqa/pull/468/commits/fd836a4d9c7a9288f537a1898015fafc1eec289f). I didn't add back the first line of that comment, though, because it seems unnecessary. The other services don't have command arguments either.

.github/actions/cache_deps/action.yml
annehaley: Ok, I like that design better. Changed in [91a7272](https://github.com/OpenImaging/miqa/pull/468/commits/91a7272d9bf1ef8fc3e2423aa84e4edb83ab7ce1)

dchiquito (approved):

PR #478, opened 2022-04-29
Release fixes 2

### LINKED ISSUE #470: Global import / export does not do anything

### LINKED ISSUE #472: Multi-project import and export
Add two new buttons on the interface, one for import one for export. When clicked, it would be equivalent to clicking import or export for each project the user has access to.

### LINKED ISSUE #473: Use WSGI server in production - we  need to use gunicorn
https://github.com/OpenImaging/miqa/blob/0246c2418d8707c6eeecb5a28e08351be78d999d/prod/docker-compose.yml#L31-L32 currently reads:
```
    # TODO install and use a wsgi server
    command: ["./manage.py", "runserver", "0.0.0.0:8000"]
```

We should use a real WSGI server to run the application in production.

annehaley:
Resolves #472. Resolves #470. Resolves #473.


prod/docker-compose.yml
```diff
@@ -28,8 +28,7 @@ services:
         DJANGO_SECRET_KEY: ${DJANGO_SECRET_KEY}
         VUE_APP_OAUTH_API_ROOT: ${VUE_APP_OAUTH_API_ROOT}
 
-    # TODO install and use a wsgi server
-    command: ["./manage.py", "runserver", "0.0.0.0:8000"]
+    command: ["gunicorn", "-k", "gthread", "--threads", "8", "--bind", "0.0.0.0:8000", "miqa.wsgi"]
```
dchiquito, line 31: Do you not need to install gunicorn?

dchiquito (approved):

prod/docker-compose.yml
annehaley: It's already included in `setup.py`

client/src/components/KeyboardShortcutDialog.vue
```diff
@@ -21,12 +21,11 @@ export default defineComponent({
           ['z', 'x'],
         ],
       ],
-      ['Decrease/increase window', [['-', '=']]],
+      ['Decrease/increase window width', [['-', '=']]],
       ['Decrease/increase window level', [['[', ']']]],
       ['Toggle fullscreen', [['e', 'd', 'c']]],
       ['Mark as usable/usable extra/questionable/unusable', [['u', 'i', 'o', 'p']]],
       ['Place crosshairs at location', [['click']]],
-      ['Changing window width & level', [['click + dragging']]],
```
zachmullen, line 29: We're removing this control?

client/src/components/VtkViewer.vue
```diff
@@ -159,6 +160,9 @@ export default {
       });
       this.resizeObserver.observe(this.$refs.viewer);
       this.view.getInteractor().onLeftButtonPress((event) => this.placeCrosshairs(event));
+      this.view.getInteractor().getInteractorStyle().getMouseManipulators().forEach(
+        (manipulator) => manipulator.setDragEnabled(false),
```
zachmullen:
```suggestion
        (manipulator) => { manipulator.setDragEnabled(false); }, 
```

client/src/django.ts
```diff
@@ -17,10 +17,6 @@ interface Paginated<T> {
 const apiClient = axios.create({ baseURL: API_URL });
 let s3ffClient;
 
-apiClient.interceptors.response.use(null, (error) => {
-  const msg = error?.response?.data?.detail || 'No response from server';
-  throw new Error(msg);
-});
```
zachmullen, line 23: Was this simply unnecessary?

miqa/core/rest/global_settings.py
```diff
@@ -25,7 +25,7 @@ def get_object(self):
         return GlobalSettings.load()
 
     def get_permissions(self):
-        if self.request.method == 'GET':
+        if self.request.method != 'PUT':
```
zachmullen: It seems safer to explicitly enumerate the conditions under which we are more permissive.

miqa/core/rest/permissions.py
```diff
@@ -37,7 +37,7 @@ def _wrapped_view(viewset, *args, **wrapped_view_kwargs):
             user_perms_on_project = get_perms(user, project)
             review_perm = has_review_perm(user_perms_on_project)
             read_perm = has_read_perm(user_perms_on_project)
-            error_response = Response(status=status.HTTP_401_UNAUTHORIZED)
+            error_response = Response(status=status.HTTP_403_FORBIDDEN)
```
zachmullen: Why construct this in advance? Can we just do it on line 47 when we know we need to return it?

zachmullen (changes requested):

client/src/django.ts
annehaley: I moved it to the bottom of the file so that I could reference `djangoClient.logout()` without a lint error

zachmullen: Oh I see, it's  still got this logic inside in addition to the new logic. Sorry for the noise.

client/src/components/KeyboardShortcutDialog.vue
annehaley: This is something I want to ask the team about before merging. The window width/level control via the vtk interactor does not affect the values of window & level shown in the sliders, and I couldn't figure out how to make this two-way. If we can figure that out, we can add it back in.

zachmullen: I think we just need to bind it manually. I assume there's some event binding we can listen to from the VTK side; we may have to ask vtk.js devs.

miqa/core/rest/permissions.py
annehaley: Sure, changed in [a31191a](https://github.com/OpenImaging/miqa/pull/478/commits/a31191aef35795936d00ec1e22910cf8c30aa8cd)

miqa/core/rest/global_settings.py
annehaley: Good point, changed in [bd32b0b](https://github.com/OpenImaging/miqa/pull/478/commits/bd32b0b660dc8fde2237726198aa787c6b40b09e)

client/src/components/KeyboardShortcutDialog.vue
annehaley: I used @floryst's advice to make the changes in [0422c1a](https://github.com/OpenImaging/miqa/pull/478/commits/0422c1a5d2371be5cfd3960dd170ab9b74820b8b), wherein I put the functionality (and keyboard shortcuts row) back in.

zachmullen (approved):

PR #486, opened 2022-05-09
Fetch the git version tag and display it in the navbar tooltip

annehaley:


miqa/core/rest/other_endpoints.py
```diff
@@ -6,6 +8,11 @@
 
 from miqa.core.models.scan_decision import ArtifactState, default_identified_artifacts
 
+MIQA_VERSION = subprocess.run(
+    ['git', 'describe', '--tags'],
```
aashish24, line 12: @annehaley this would always generate the last tag available on the master branch. Wouldn't it be better if we have the SHA that was used to build the image?

annehaley: Added in [e2ec803](https://github.com/OpenImaging/miqa/pull/486/commits/e2ec803b034eac53c42e80977447951e23261029)

PR #487, opened 2022-05-10
Add pyppeteer tests for imports and exports

annehaley:


miqa/core/tests/pyppeteer/test_import_export.py
```diff
@@ -0,0 +1,100 @@
...
+
+async def get_current_num_scans(page):
+    return len(await page.xpath('//ul[contains(@class, "scans")]/li'))
+
+
+async def edit_import_export_paths(page):
+    # Edit import path
+    import_path_input = await page.waitForXPath('//input[contains(@name, "import-path")]')
+    await import_path_input.click({'clickCount': 3})  # highlight the text to overwrite it
+    await import_path_input.type('samples/demo_project.csv')
```
dchiquito: I think it's worth having `import_path` and `export_path` arguments to this function. If someone were reading the test, they would need to dig into this method to find what exactly is being put into the import/export fields.

miqa/core/tests/pyppeteer/test_import_export.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Edit export path
+    import_path_input = await page.waitForXPath('//input[contains(@name, "export-path")]')
+    await import_path_input.click({'clickCount': 3})  # highlight the text to overwrite it
+    await import_path_input.type('samples/demo_project_export.csv')
+    await page.waitFor(1_000)
+
+    # Click save
+    await (await page.waitForXPath('//button[span[.=" Save "]][@type="submit"]')).click()
+    await page.waitFor(1_000)
```
dchiquito, line 22: If there's a save confirmation notification, it would be good to check for that here.

miqa/core/tests/pyppeteer/test_import_export.py
```diff
@@ -0,0 +1,100 @@
...
+async def perform_import(page):
+    # Click import button, and again in the modal confirmation
+    await (await page.waitForXPath('//button[span[.=" Import "]]')).click()
+    await page.waitFor(1_000)
+    await (
+        await page.waitForXPath(
+            '//div[contains(@class, "v-card__actions")]//button[span[.=" Import "]]'
+        )
+    ).click()
+    await page.waitFor(6_000)
```
dchiquito, line 36: Generally speaking, calling `waitFor(...)` immediately before `waitForXPath(...)` doesn't accomplish very much. [waitForXPath](https://puppeteer.github.io/puppeteer/docs/puppeteer.page.waitforxpath/) already waits for a while (30 seconds by default, I think? It's configurable when the `page` fixture is set up), and you can set an arbitrary timeout when calling waitForXPath. If it really takes more than 30s but less than 36s for the element to appear, I think it's probably better to just pass the `timeout` parameter to waitForXPath.

That said, there are exceptions when it does make sense to do exactly this. If you had a search widget with some search results already present, then you do a new search, you would need to wait a moment with `waitFor` for the old results to be cleared away, then wait for the new results with `waitForXPath`. This doesn't seem like one of those times, though.

miqa/core/tests/pyppeteer/test_import_export.py
```diff
@@ -0,0 +1,100 @@
...
+    creator = user_factory()
+    project = project_factory(creator=creator, name='Demo Project')
+    project.update_group('tier_2_reviewer', [creator])
+
+    await log_in(creator)
+    # Click on the project in the project selector
+    await (
+        await page.waitForXPath(f'//div[contains(@class, "col")][contains(.,"{project.name}")]')
+    ).click()
+    await page.waitFor(1_000)
```
dchiquito, line 69: For future reference:
If you end up writing enough end to end tests that you have multiple tests that use the same page/widget to test completely different functionality, it becomes worth it to create modules/classes that represent a page/component in the app and keep all the helpers on that page helper rather than in the test method.

Setting that up now to abstract away 4 lines (2 statements?) of code seems like overkill, but it's something to think about planning for as you write more tests.

dchiquito (approved):
:+1: Looks good to me, just a few minor tweaks I'm sure you can take care of.

Completely unrelated thought: If Jim has instructions to reproduce the log in bug, you can probably write a test that reproduces it reliably, and then also be assured that the bug is really resolved  when the test no longer fails. https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings can be used to edit the session expiration settings.

:wave:

miqa/core/tests/pyppeteer/test_import_export.py
annehaley: Changed in [a918a10](https://github.com/OpenImaging/miqa/pull/487/commits/a918a1065eb966c36b2c2a9b13babb9dcb1dc196)

miqa/core/tests/pyppeteer/test_import_export.py
annehaley: Changed in [bbd595e](https://github.com/OpenImaging/miqa/pull/487/commits/bbd595edc232fb0863ce00b8d683c42743a65e7b)

PR #494, opened 2022-05-18
Use heartbeat to log users out even if tab is closed

### LINKED ISSUE #485: Session expiration does not seem to work as expected
@jimklo could you provide steps to reproduce

zachmullen:
Fixes #485.

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -20,7 +20,8 @@ import vMousetrap from './vue-utilities/v-mousetrap';
 import snackbarService from './vue-utilities/snackbar-service';
 import promptService from './vue-utilities/prompt-service';
 
-import djangoRest from './django';
+import djangoRest, { oauthClient } from './django';
+import { setupHeartbeat } from './heartbeat';
 
 Vue.use(Vuetify);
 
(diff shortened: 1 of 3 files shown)


client/src/main.ts
```diff
@@ -44,21 +45,24 @@ Sentry.init({
   dsn: process.env.VUE_APP_SENTRY_DSN,
 });
 
-djangoRest.restoreLogin(store).then(async () => {
+(async () => {
+  // If user closes the tab, we want them to be logged out if they return to the page
```
annehaley, line 49: This is only the case if they have the tab closed for at most ten minutes, right? Is it desired that their token is immediately invalidated once they close the tab?

zachmullen: I wanted to make this strictly less than the server-side auth duration (which is 30 minutes). Note that this code only invalidates the _oauth_ token, not the server-side login credentials. So if they return to the tab and their server-side token is still in the valid 30-minute window, they will go through the redirect loop to reacquire a new oauth token, but won't have to actually log in. To the user, it will just appear to take slightly longer to load the page after that 10 minutes but before 30.

annehaley: Ok, that makes sense. Thanks for explaining.

annehaley (approved):

PR #497, opened 2022-05-18
Neurology toggle

### LINKED ISSUE #279: Add Project setting to toggle between neurology and radiology orientations
Currently, the definition of view orientations are hard-coded in the frontend as the expected axis orientations for neurology, so radiology scans appear flipped. We should add a toggle to the Project settings that can tell the `VtkViewer` which orientations to use.

annehaley:
Resolves #279 
@curtislisle could you confirm that the RAS specification in `constants.js` is correct?


miqa/core/models/project.py
```diff
@@ -39,6 +39,13 @@ class Project(TimeStampedModel, models.Model):
     archived = models.BooleanField(default=False)
     import_path = models.CharField(max_length=500, blank=True)
     export_path = models.CharField(max_length=500, blank=True)
+    neurology_orientation = models.BooleanField(
```
zachmullen: Change to `CharField` with `TextChoices` that directly indicate the anatomical orientation.

zachmullen (changes requested):

miqa/core/models/project.py
annehaley: Changed in [9ebda65](https://github.com/OpenImaging/miqa/pull/497/commits/9ebda653a0f3119795945f8e5e62c633878bd74f)

client/src/vtk/viewManager.js
```diff
@@ -61,7 +61,7 @@ function getView(proxyManager, viewType) {
       .forEach((s) => proxyManager.getRepresentation(s, view));
 
     // Update orientation
-    const { axis, orientation, viewUp } = VIEW_ORIENTATIONS[name];
+    const { axis, orientation, viewUp } = VIEW_ORIENTATIONS.LPS[name];
```
zachmullen, line 67: Why is this hardcoded as LPS?

miqa/core/models/project.py
```diff
@@ -39,6 +39,11 @@ class Project(TimeStampedModel, models.Model):
     archived = models.BooleanField(default=False)
     import_path = models.CharField(max_length=500, blank=True)
     export_path = models.CharField(max_length=500, blank=True)
+    anatomy_orientation = models.CharField(
+        choices=[('LPS', 'LPS'), ('RAS', 'RAS')],
```
zachmullen: Let's use `TextChoices` for this: https://docs.djangoproject.com/en/4.0/ref/models/fields/#enumeration-types

client/src/components/ProjectSettings.vue
```diff
@@ -53,12 +55,15 @@ export default defineComponent({
           await djangoRest.setGlobalSettings({
             import_path: importPath.value.trim(),
             export_path: exportPath.value.trim(),
+            anatomy_orientation: undefined,
```
zachmullen: Since the server-side logic is expecting this key not to be set, I think the more straightforward expression of that is to just omit it here.
```suggestion
```

The "right" thing to do for this request would be to support a partial update of this object via an HTTP PATCH, but that's out of scope here, and not really that important in this context.

zachmullen (changes requested):

client/src/vtk/viewManager.js
annehaley: This is simply in the constructor as a default so that we can create the view objects in the store upon initial load (before `currentProject` is guaranteed to be defined). Once the scans are loaded in (at which point `currentProject` IS guaranteed to be defined), the `VtkViewer` initializes its camera referencing the project's chosen orientation.

miqa/core/models/project.py
annehaley: Changed in [ba51aeb](https://github.com/OpenImaging/miqa/pull/497/commits/ba51aeb17481047426c09c5a11c71f487536c2e1)

client/src/components/ProjectSettings.vue
annehaley: Changed in [463db2b](https://github.com/OpenImaging/miqa/pull/497/commits/463db2b8c82c4e83c5e6dc928baec4287c22e3b6)

client/src/vtk/viewManager.js
zachmullen: Ok, could we write that in a comment here?

annehaley: Yep! Added in [1985f0a](https://github.com/OpenImaging/miqa/pull/497/commits/1985f0ae6afdd76cdd260c1eacdd1709c77393fc)

zachmullen (approved):

PR #514, opened 2022-06-01
Pre release fixes

annehaley:


miqa/core/rest/project.py
```diff
@@ -56,13 +56,23 @@ def get_default_email_recipients(self, obj):
...
-        fields = ['total_experiments', 'total_scans', 'my_project_role', 'scan_states']
+        fields = [
+            'total_experiments',
+            'total_scans',
+            'my_project_role',
+            'scan_states',
+            'project_id',
+        ]
 
+    project_id = serializers.SerializerMethodField('get_project_id')
```
zachmullen:
```suggestion
    project_id = serializers.UUIDField(source='id')
```

Then you can get rid of the method below.

miqa/core/tests/test_import.py
```diff
@@ -34,7 +34,7 @@ def test_import_empty_csv(tmp_path, user, project_factory, user_api_client):
 def test_import_csv(tmp_path, user, project_factory, sample_scans, user_api_client):
     csv_file = str(tmp_path / 'import.csv')
     with open(csv_file, 'w') as fd:
-        output, _writer = generate_import_csv([sample_scans[0]])
+        output, _writer = generate_import_csv([scan for scan in sample_scans if 'ucsd' in scan[0]])
```
zachmullen, line 37: Why the conditional here? Shouldn't this test always have the same value?

zachmullen (changes requested):

miqa/core/tests/test_import.py
annehaley: After we added the logic in the import method that checks whether the value in the project name column matches the project for which you are performing the import, this test became non-deterministic. The `sample_scans` fixture reutrns the scans in our `samples` directory. Most often, that would result in the ucsd scan followed by the ohsu scan. I had tried to make these tests pass an import for a project called "ucsd" by only using the first element in the `sample_scans` fixture. But sometimes the tests would receive `sample_scans` as the ohsu scan followed by the ucsd scan (and thus the test would fail because the first scan would be labelled for ohsu instead of ucsd). I needed to make this deterministic, so instead of using the _first element_ in `sample_scans`, I filter for _any elements that match the project name_ for which I will be performing the import.

zachmullen: I see, thanks for the explanation.

zachmullen (approved):

PR #533, opened 2022-07-01
Python client

annehaley:
This client implements the classes/methods planned here: https://docs.google.com/document/d/1hdwDVFfU5yGSEV_BHjUSxrH0XyQjkkO2qxEcAtQAh34/edit?usp=sharing

I'm planning on adding examples of usage in another PR. This one is focused on the design of these classes.

I'm not sure why the Cloudflare build step is consistently failing...


python_client/MIQA.py
```diff
@@ -0,0 +1,130 @@
+from getpass import getpass
+import requests
+from s3_file_field_client import S3FileFieldClient
+
+
+from project import Project
```
zachmullen:
```suggestion
from .project import Project
```

python_client/decision.py
```diff
@@ -0,0 +1,32 @@
+class ScanDecision:
```
zachmullen: Let's make these objects `dataclass`es: https://docs.python.org/3/library/dataclasses.html

python_client/MIQA.py
```diff
@@ -0,0 +1,130 @@
...
+            POST=True,
+            body={
+                'name': name,
+            },
+        )
+        new_project = Project(**dict(response, MIQA=self))
+        self.projects.append(new_project)
+        return new_project
+
+    def list_all_objects(self, indent=0):
```
zachmullen: If we want to keep these, let's change the names from `list_` to `print_`

python_client/experiment.py
```diff
@@ -0,0 +1,86 @@
...
+        self.note = note
+
+    def get_scan_by_id(self, id: str):
+        if self.scans:
+            matches = [scan for scan in self.scans if scan.id == id]
+            if len(matches) == 1:
+                return matches[0]
+        try:
+            response = self.project.MIQA.make_request(f"scans/{id}", GET=True)
+        except Exception:
```
zachmullen: `except Exception` is usually a bad practice -- let's narrow the scope of this catch type to only catch the exact type of known failure modes for which we'd want to return `None`. In this case, it would be an HTTP 404.

zachmullen (changes requested):

python_client/MIQA.py
annehaley: If I use relative imports here, I get `ImportError: attempted relative import with no known parent package`

zachmullen: Ah, yes, we'll want to make the python client a proper package. Add an `__init__.py` to do so.

We should also consider making it a publishable package by moving the code down into its own directory and adding a `setup.py` so that it can be `pip` installable.

python_client/experiment.py
annehaley: Changed in [1f9bf57](https://github.com/OpenImaging/miqa/pull/533/commits/1f9bf5786fe0af1bc19bed1b7d0128275ee56144)

python_client/MIQA.py
annehaley: Changed in [413dfdb](https://github.com/OpenImaging/miqa/pull/533/commits/413dfdbe699db290e369f3d8f44bfdd58b8378e6)

python_client/decision.py
annehaley: Changed in [9ad531d](https://github.com/OpenImaging/miqa/pull/533/commits/9ad531d60498d42b7330a3cfc5384f7d12aa6b9f)

python_client/MIQA.py
annehaley: Changed in [71e76c3](https://github.com/OpenImaging/miqa/pull/533/commits/71e76c37163f0c4c61a3a9aead7a4496f89558de). Do you know why the Cloudflare build keeps failing?

python_client/setup.py
```diff
@@ -0,0 +1,40 @@
...
+        "asgiref",
+        "certifi",
+        "Django>3.2,<4.0",
+        "django-s3-file-field-client==0.3.0",
+        "djangorestframework>3",
+        "idna",
+        "pytz",
+        "requests",
+        "sqlparse",
+        "urllib3",
```
zachmullen: Reduce this to just the (unpinned) set of libraries that we directly depend on, which looks like just `requests` and `django-s3-file-field`.

python_client/src/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+            if len(matches) == 1:
+                return matches[0]
+        api_path = f"scans/{id}"
+        response = requests.get(
+            f"{self.project.MIQA.url}/{api_path}",
+            headers=self.project.MIQA.headers,
+        )
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
```
zachmullen: Rather than `try / except` you can check `response.ok`

Also check the status code in this specific case, since we only want to return `None` if it's HTTP 404.

python_client/src/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+            f"{self.project.MIQA.url}/{api_path}",
+            headers=self.project.MIQA.headers,
+            json={
+                'note': note,
+            },
+        )
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            raise MIQAAPIError(f'Request failed: {response.json()}')
```
zachmullen: Let's not wrap RequestExceptions, just raise them.

annehaley: Changed in [bd731b1](https://github.com/OpenImaging/miqa/pull/533/commits/bd731b13a808d215e55f72dcf4911101d5c6d6d6)

python_client/src/experiment.py
annehaley: Changed in [bd731b1](https://github.com/OpenImaging/miqa/pull/533/commits/bd731b13a808d215e55f72dcf4911101d5c6d6d6)

python_client/setup.py
annehaley: Changed in [426a913](https://github.com/OpenImaging/miqa/pull/533/commits/426a9131d375705a0b135b63df6920d1a73bcf76)

zachmullen (approved):

PR #537, opened 2022-07-13
Python client docs

annehaley:


python_client/README.md
```diff
@@ -0,0 +1,75 @@
+# MIQA Python Client
+This simple python library is intended to allow users of a MIQA instance to interact with the MIQA API programmatically and perform tasks they may not be able to accomplish in the web interface.
+
+The structure of classes in this library closely imitates the structure of models in MIQA, so it is recommended that a user understands this structure before proceeding to use the python client. The documentation explaining this structure can be found at https://openimaging.github.io/miqa/projects/.
+
+In summary, Projects are the top-level object that act as organizational buckets for experiments. Experiments are collections of scans that may be related in some way. Scans are collections of related 3D medical images. Each scan must have at least one image, or Frame, associated with it. A scan may have multiple Frames representing time steps or some other fourth dimension to the images collected. The purpose of MIQA is to mark scans as usable, unusable, or otherwise. Thus, a Scan Decision is the object that stores those determinations. A Scan can have zero or more Scan Decisions.
```
aashish24: nit pick - A project is the top level object that acts as or Projects are top level objects that act as

aashish24: Also --> The purpose of MIQA is to annotate the quality of scans that are currently categorized as usable, unusable, or otherwise. Thus, a Scan Decision is an object that stores those determinations. A Scan can have zero or more Scan Decisions.

python_client/example.py
```diff
@@ -0,0 +1,143 @@
...
+    instance.artifact_options,
+)
+
+# --- Project ---
+# if we still haven't called instance.get_all_objects()
+# this will request only the target project to load into instance.projects
+# this may be useful if you only want certain projects loaded
+project_a = instance.get_project_by_id('6c3ecc33-ee41-45a8-a31a-230ca9b574f0')
+project_a.print_all_objects()
+# You can create projects if you are a privileged user in this MIQA instance
```
aashish24, line 37: what does it mean to be a privileged user?

python_client/src/decision.py
```diff
@@ -10,7 +10,7 @@
 class ScanDecision:
     """
     Attributes:
-      id, name, decision, creator, creation_time, note,
+      id, decision, creator, created, note,
```
aashish24, line 13: minor comment: created is a verb but everything else is noun

aashish24:
Looking very good, thanks for putting these detailed documents. I just had very few minor comments

annehaley:
> Looking very good, thanks for putting these detailed documents. I just had very few minor comments

I don't see your comments. Did you write them as a review on this PR?

dzenanz (approved):

aashish24:
@annehaley do you see it now?

annehaley:
> @annehaley do you see it now?
Yes, thank you!

python_client/src/decision.py
annehaley: I would've left it at `creation_time`, which makes more sense to me, but the field in the DB is called `created` so that is what the API calls the field. I figured we should prioritize congruence with the API responses to avoid confusion.

python_client/example.py
annehaley: Added more explanation in [2c7e3fe](https://github.com/OpenImaging/miqa/pull/537/commits/2c7e3fe2cb2f999802557f40ee728feaa54177b3)

python_client/README.md
annehaley: Changed in [2c7e3fe](https://github.com/OpenImaging/miqa/pull/537/commits/2c7e3fe2cb2f999802557f40ee728feaa54177b3)

PR #538, opened 2022-07-14
UI fixes

### LINKED ISSUE #534: Kilian's suggested UI fixes
- scan states always show "unreviewed" in tooltips and donut chart
- scan decision code should come with a tooltip showing the full text of the decision type in the previous decisions box
- cursor should change to pointer finger when hovering over the scan link
- emails should include scan link, subject id, and session id
- screenshots attached to emails should include crosshairs if the main view has crosshairs enabled
- window width & level widget redesign (#214)

annehaley:
Resolves #534 
I could not replicate the first item, wherein the states of scans with existing ScanDecisions show up as unreviewed.

### DIFF
--- a/web_client/src/components/ControlPanel.vue
+++ b/web_client/src/components/ControlPanel.vue
@@ -204,120 +204,108 @@ export default {
             height="100%"
             elevation="3"
           >
-            <v-container fluid>
-              <v-flex
-                class="d-flex justify-space-between"
+            <div style="display: flex">
+              <v-container>
+                <v-row no-gutters>
+                  <v-col cols="3">
+                    Project:
+                  </v-col>
+                  <v-col cols="9">
+                    {{ currentViewData.projectName }}
+                  </v-col>
+                </v-row>
+                <v-row no-gutters>
+                  <v-col cols="3">
+                    Experiment:
+                  </v-col>
+                  <v-col cols="9">
+                    <div v-if="lockOwner">
+                      <div>
+                        {{ currentViewData.experimentName }}
+                        <UserAvatar
+                          v-if="lockOwner"
+                          :target-user="lockOwner"
+                          as-editor
+                        />
+                      </div>
+                    </div>
+                  </v-col>
+                </v-row>
+                <v-row no-gutters>
+                  <v-col cols="3">
+                    Subject:
+                  </v-col>
+                  <v-col cols="9">
+                    <b>{{ currentViewData.scanSubject || 'None' }}</b>
+                  </v-col>
+                </v-row>
+                <v-row no-gutters>
+                  <v-col cols="3">
+                    Session:
+                  </v-col>
+                  <v-col cols="9">
+                    <b>{{ currentViewData.scanSession || 'None' }}</b>
+                  </v-col>
+                </v-row>
+              </v-container>
+              <div
+                v-if="scanCachedPercentage < 1 && scanCachedPercentage > 0"
+                class="ma-10"
               >
-                <div
-                  class="d-flex"
-                  style="flex-direction:column; row-gap: 10px;"
-                >
-                  <span>Project:</span>
-                  <span>Experiment:</span>
-                </div>
-                <div
-                  rows="2"
-                  class="py-3"
-                  style="text-align: center; height: 70px"
-                >
-                  <div v-if="scanCachedPercentage < 1 && scanCachedPercentage > 0">
-                    <v-progress-circular
-                      :value="scanCachedPercentage * 100"
-                      color="blue"
-                    />
-                    <div> Loading... </div>
-                  </div>
-                </div>
-                <div
-                  v-if="lockOwner"
-                  class="grey--text d-flex"
-                  style="text-align: right; flex-direction:column; row-gap: 5px;"
-                >
-                  <span>{{ currentViewData.projectName }}</span>
-                  <div>
-                    <UserAvatar
-                      v-if="lockOwner"
-                      :target-user="lockOwner"
-                      as-editor
-                    />
-                    {{ currentViewData.experimentName }}
-                  </div>
-                </div>
-              </v-flex>
-              <v-row
-                v-if="currentViewData.scanSession || currentViewData.scanSubject"
-                no-gutters
+                <v-progress-circular
+                  :value="scanCachedPercentage * 100"
+                  color="blue"
+                />
+                <div> Loading... </div>
+              </div>
+            </div>
+            <v-textarea
+              v-model="currentViewData.experimentNote"
+              filled
+              :disabled="!experimentIsEditable"
+              no-resize
+              height="60px"
+              hide-details
+              name="input-experiment-notes"
+              label="Experiment Notes"
+              placeholder="There are no notes on this experiment."
+              style="margin-top: 15px;"
+              class="mx-3"
+              @input="handleExperimentNoteChange"
+            />
+            <v-row no-gutters>
+              <v-col
+                :class="newExperimentNote.length > 0 ? 'blue--text' : 'grey--text'"
+                style="text-align: right"
+                class="px-3"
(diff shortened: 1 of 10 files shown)


web_client/src/components/EmailDialog.vue
```diff
@@ -84,10 +84,19 @@ export default defineComponent({
       if (this.user) {
         this.cc.push(this.user.email);
       }
+
+      console.log(this.currentScan);
```
zachmullen:
```suggestion
      console.log(this.currentScan);
```

web_client/src/components/VtkViewer.vue
```diff
@@ -249,33 +248,56 @@ export default {
...
-      Vue.nextTick(() => {
-        fill2DView(this.view);
-        setTimeout(() => {
-          this.resized = true;
-        });
-      });
+      setTimeout(() => {
+        this.$refs.viewer.style.width = 'inherit';
+        this.$refs.viewer.style.width = `${this.$refs.viewer.clientWidth - 3}px`;
+      }, 100);
```
zachmullen, line 300: Why setTimeout with 100ms? If `nextTick` isn't sufficient, this probably merits a comment explaining why.

zachmullen (changes requested):

web_client/src/components/VtkViewer.vue
annehaley: I wish I could explain why; it had me stumped for a while. But I don't know why `Vue.nextTick` didn't work. It fired, but failed to resize the image.

web_client/src/components/EmailDialog.vue
annehaley: Changed in [5fbde61](https://github.com/OpenImaging/miqa/pull/538/commits/5fbde613f889ef3cde2eb59ce38e845ca95c696c)

web_client/src/components/VtkViewer.vue
zachmullen: Ah ok. Maybe we need to listen to the viewer element's native resize event and then trigger this as a callback?

annehaley: We have a resizeObserver which observes `this.$refs.viewer`. It fires as well, but also fails to resize the image. The only way I found to work was with a slight delay, doing something that would cause the resizeObserver to trigger

kipohl:
@annehaley so UN is unusable - anything else you want me to check on ?

annehaley:
>  anything else you want me to check on ?

@kipohl  What do you think of this control panel layout? You asked for a change in the alignment of scan name, scan subject, scan session, and frame number. I moved some things around to re-fit the three sub-panels:
image

kipohl:
a) Scan Subject and Scan Session should be labelled Subject and Session 
b) Given that those are not scan specific I think it would be most intuitive to move those to the left pannel and move Window width and level to the middle panel  (also those are far down - should be closer to the window itself) - so middle pannel would be scan , frame , window level /width . If you are running out of space put the scan on the left panel.
c) Window presets could stay on the left pannel as again those are not scan specific or move it to the middle , 
d) Displaying cross hair (and next to it)  should be again on left panel

annehaley:
@kipohl How about this?
image

kipohl:
much , much better 
left panel: Order Project, Subject , Experiment , Session 
middle panel: create a box with just window stuff so you can shorten text to width, length , preset  - move locks further to the right

annehaley:
I changed the appearance of the window widget: 
image

But can I ask why the reordering of labels in the left panel? In Curt's use case, both the subject and session may change per experiment and per scan. So putting subject between project and experiment may be confusing.

kipohl:
better - what about just having window appear once to reduce text

kipohl:
e.g. as header of box - if you put it on the bound it should not take up too much more vertical space

kipohl:
something like Window Setting

kipohl:
also do not know why preset options has to appear twice - again you can save here some vertical space

kipohl:
in left panel would not align second column on the right but on left -

kipohl:
labels and values are too far away from each other so that it makes harder to read

curtislisle:
I like the new widget. I suggest  we put only the word “window” on the top slider and only “level” on the bottom slider.

annehaley:
With the new changes, the control panel now looks like this:
image

kipohl:
I find it much cleaner - one question - is the window preset applied to an individual scan or once it is said it is applied to the next scan too ? If the second, then window preset should be moved to the left panel - maybe then include a box around cross hair, ... , window preset with the label settings

annehaley:
> I find it much cleaner - one question - is the window preset applied to an individual scan or once it is said it is applied to the next scan too ? If the second, then window preset should be moved to the left panel - maybe then include a box around cross hair, ... , window preset with the label settings

Any changes to the window settings can be carried over to the next scan by toggling the lock on those settings. Applying a preset changes the width and level values. Whether those new values carry over is dependent on the lock state. I think we should keep the preset options with the rest of the window widget.

kipohl:
To clarify - do you mean frame by scan ? If so this would be the wrong behavior in general - locks should only be in place per frame not scan

annehaley:
This can be up to the user. If they unlock the window width & level after finishing a scan, navigating to the next one will reset those values. Keeping the window locked between frames is doable, or keeping the window locked between scans, or between experiments.

kipohl:
can you then have a switch in the left panel stating that window lock should be kept across scans ? 
I am asking as for for my use case the default logic makes no sense but I can think of others where it would

web_client/src/components/ControlPanel.vue
```diff
@@ -204,92 +204,108 @@ export default {
...
+              v-model="currentViewData.experimentNote"
+              filled
+              :disabled="!experimentIsEditable"
+              no-resize
+              height="60px"
+              hide-details
+              name="input-experiment-notes"
+              label="Experiment Notes"
+              placeholder="There are no notes on this experiment."
+              style="margin-top: 15px;"
```
zachmullen: Why not just apply another vuetify class like `mt-2`?

web_client/src/components/ControlPanel.vue
```diff
@@ -204,92 +204,108 @@ export default {
...
+              class="mx-3"
+              @input="handleExperimentNoteChange"
+            />
+            <v-row no-gutters>
+              <v-col
+                :class="newExperimentNote.length > 0 ? 'blue--text' : 'grey--text'"
+                style="text-align: right"
```
zachmullen: Isn't there a vuetify class for this as well?

web_client/src/components/ControlPanel.vue
```diff
@@ -204,92 +204,108 @@ export default {
             height="100%"
             elevation="3"
           >
-            <v-container fluid>
-              <v-flex
-                class="d-flex justify-space-between"
+            <div style="display: flex">
```
zachmullen: Why the manual style tag here? Isn't flexbox stuff handled for us with vuetify layout utilities?

annehaley: Changed in [f47f9d9](https://github.com/OpenImaging/miqa/pull/538/commits/f47f9d99cc6d01d74b68bc856401f459866cc609)

web_client/src/components/ControlPanel.vue
annehaley: Changed in [f47f9d9](https://github.com/OpenImaging/miqa/pull/538/commits/f47f9d99cc6d01d74b68bc856401f459866cc609)

web_client/src/components/ControlPanel.vue
annehaley: Changed in [f47f9d9](https://github.com/OpenImaging/miqa/pull/538/commits/f47f9d99cc6d01d74b68bc856401f459866cc609)

PR #539, opened 2022-07-15
Update NN weights with April work

dzenanz:


annehaley (approved):
These changes look good. Do you have any data on how the inferences improved with this training?

dzenanz:
Improvement is not significant. The training is a bit more consistent.
February-April change
Blue is old, red is new.

annehaley:
Is there a known reason for the spike dip in R2 at ~700k steps?

dzenanz:
That is first iteration of fold 3. Usually, the first few iterations have bad accuracy.

For each fold the best point is what we keep. And the convergence towards a certain value gives confidence.

PR #540, opened 2022-07-18
Add weight files from training on T1/T2/PD mixture

dzenanz:
We are switching to combining T1, T2 and PD images.


dzenanz:
This should have a noticeable improvement. I think we need to change somewhere which weight file is used by default. I think we are currently using `miqaT1-val0.pth`. We should change this to `miqaMix-val0.pth`.

annehaley (changes requested):
You can simply change line 29 in learning/evaluation_models.py. If you want you can change the name of the model in that mapping, too, but then you'd also have to update the Project model. It's easiest if you only change the filename for that entry.

dzenanz:
@annehaley is this the right way to do it?

annehaley:
Changing the `evaluation_model_mapping` in `project.py` requires a new migration to apply this change to existing projects in the database. It might be easier to only change learning/evaluation_models.py

dzenanz:
Then we have the wrong name, `T1` instead of `Mix`. How much of an overhead is to create/apply a django migration? We could discuss this during Thursday meeting.

annehaley:
I was wrong about the migration. I forgot that we do content validation of the `evaluation_model_mapping` field in our `Project.clean` method instead of applying formal DB constraints on the field. So this change does not actually need a migration. 

But one change you'll need to get your tests to pass will be updating line 71 in `models/project.py`

dzenanz:
Thanks Anne, it passes checks now!

annehaley (approved):

PR #552, opened 2022-08-02
Window widget 2

### LINKED ISSUE #214: Window width & level widget redesign
The UI widgets for window width and window level were greatly simplified in the control panel redesign, and we may need to reconsider the functionality needed for the user to review scans. Do the minimum and maximum values of the slider need to be mutable by the user? Should these values be set per frame or per scan?

### LINKED ISSUE #548: Replace two window sliders (width and level) for one range slider just like Slicer

annehaley:
Resolves #214. Resolves #548.
This PR addresses:

1. Combining the window width and level sliders into one range slider, which is responsive to changes made to the window via render interaction
2. Removing the "Apply" button; presets are automatically applied when selected
3. Using a special lock mechanism that will maintain the lock state for all the frames in the current scan, current experiment, or current project
4. Moving both ends of the range simultaneously by dragging the middle of the range bar (i.e. change the level without changing the width) 

image

Here's a description of the math used to compute the values of the range:

The range shown in blue on the range slider is the `currentRange`, `[v0, v1]`.
The window width, `ww`, is `v1 - v0`. 
The window level, `wl`, is the midpoint between `v0` and `v1`.

Therefore:
```
v0 = wl - ww / 2
v1 = wl + ww / 2
ww = v1 - v0
wl = v0 + ww / 2
```


zachmullen (approved):
Code LGTM

dzenanz:
Don't forget flipping the axial view in RAS view.

annehaley:
> Don't forget flipping the axial view in RAS view.

Thanks! fixed in [7949ae4](https://github.com/OpenImaging/miqa/pull/552/commits/7949ae4a72c3a9bd7129fabeac5093e3e2e6b223)

zachmullen:
Code LGTM. Let's deploy it to the demo and see if we find any bugs playing with it there.

PR #556, opened 2022-08-04
Disable arrowkey event listener if an input is focused

annehaley:


web_client/src/components/ControlPanel.vue
```diff
@@ -72,6 +72,7 @@ export default {
     if (!this.navigateToNextIfCurrentScanNull()) {
       this.switchLock(this.experimentId);
       window.addEventListener('keydown', (event) => {
+        if (['textarea', 'input'].includes(document.activeElement.type)) return;
```
dzenanz, line 75: Does this also affect text inputs outside of the email modal dialog, such as free-form comments?

annehaley: Yes, this applies to all inputs and textareas in the DOM

dzenanz: This is worthy of mention in release notes, so that user's don't think we broke hotkeys.

PR #561, opened 2022-08-09
Bug hunt

annehaley:
This PR cleans up code areas that commonly cause Sentry/CI errors. Each commit is dedicated to one error and explains the change in the commit message.

### DIFF
--- a/web_client/src/django.ts
+++ b/web_client/src/django.ts
@@ -52,7 +52,11 @@ const djangoClient = {
     try {
       await apiClient.post('/logout/', undefined, { withCredentials: true });
     } finally {
-      await oauthClient.logout();
+      await oauthClient.logout().then(
+        async () => {
+          await oauthClient.redirectToLogin();
+        },
+      );
     }
   },
   async MIQAConfig() {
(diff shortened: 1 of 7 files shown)


web_client/src/django.ts
```diff
@@ -52,7 +52,11 @@ const djangoClient = {
     try {
       await apiClient.post('/logout/', undefined, { withCredentials: true });
     } finally {
-      await oauthClient.logout();
+      await oauthClient.logout().then(
+        async () => {
+          await oauthClient.redirectToLogin();
```
zachmullen, line 57: Are we sure this doesn't cause an infinite loop? I remember when refactoring this before that that would happen under certain conditions.

annehaley: What we had there before causing the infinite loop was `window.location.reload`. In my testing, this did not loop.

zachmullen: Ok, we can deploy this and I'll do some testing on the demo site to make sure no loop occurs.

zachmullen (approved):

PR #563, opened 2022-08-15
Bug hunt 2

annehaley:


miqa/core/tests/pyppeteer/test_scan_upload.py
```diff
@@ -63,6 +63,6 @@ async def test_upload_scans(
     await scans[0].click()
 
     assert await page.waitForXPath(
-        '//div[contains(@class, "col")][contains(., "Demo Project")]'
-        '/following::div[contains(@class, "col")][contains(., "Test Experiment")]'
+        '//div[contains(@class, "d-flex")][contains(., "Demo Project")]'
```
zachmullen: It's probably a better practice to put a custom class on this rather than strictly depending on a `d-flex`-classed element. E.g. `class="miqa-project-title"`

miqa/core/tests/pyppeteer/test_scan_upload.py
```diff
@@ -63,6 +63,6 @@ async def test_upload_scans(
...
+        '/following::div/div[contains(., "Test Experiment")]'
```
zachmullen, line 67: Same here; rather than depending on a specific nesting of the DOM, we should just add a special class to the element where this text is expected.

web_client/src/components/ControlPanel.vue
```diff
@@ -206,52 +206,64 @@ export default {
             elevation="3"
           >
             <div class="d-flex">
-              <v-container>
-                <v-row no-gutters>
-                  <v-col cols="3">
+              <div
+                class="d-flex px-5 py-3"
+                style="flex-direction: column; flex-grow: 3"
```
zachmullen: Do Vuetify's normal flex- and layout-related components and classes not handle what we want to do here?

web_client/src/components/ControlPanel.vue
```diff
@@ -206,52 +206,64 @@ export default {
...
-                </v-row>
-              </v-container>
+                  </div>
+                </div>
+              </div>
               <div
                 v-if="scanCachedPercentage < 1 && scanCachedPercentage > 0"
-                class="ma-10"
+                class="px-6 py-8"
+                style="text-align: center"
```
zachmullen: I think there's a vuetify class for this, maybe `align-center` IIRC

web_client/src/components/DataImportExport.vue
```diff
@@ -80,7 +80,8 @@ export default defineComponent({
         } else {
           response = await djangoRest.projectExport(currentProject.value.id);
         }
-        if (response.warnings) {
+        console.log(response);
```
zachmullen:
```suggestion
```

web_client/src/components/ExperimentsView.vue
```diff
@@ -158,6 +158,9 @@ export default {
           const newExperiment = await djangoRest.createExperiment(
```
zachmullen: Could we just have `createExperiment` throw its own error if something goes wrong? It could throw a custom error type that contains relevant info.

web_client/src/components/WindowWidget.vue
```diff
@@ -53,8 +54,9 @@ export default defineComponent({
     function updateFromRange([v0, v1]) {
       if (windowLocked.value) return;
       if (v0 === currentRange.value[0] && v1 === currentRange.value[1]) return;
+      currentRange.value = [v0, v1];
       const ww = v1 - v0;
-      const wl = v0 + Math.ceil(ww / 2);
+      const wl = v0 + Math.floor(ww / 2);
```
zachmullen, line 59: Should we just do `Math.round`?

zachmullen (changes requested):
comments inline

kipohl:
@annehaley what do u want me to comment on

annehaley:
> @annehaley what do u want me to comment on

Just wanted your confirmation that this PR addresses your requests.

web_client/src/components/WindowWidget.vue
annehaley: The problem I discovered with rounding up here is that it sometimes creates the cyclical update case where the range values are updated by +1 each tick. This happened when the upper limit had any kind of decimal (which happens when the values are computed by changing the window via a click-and-drag in the render). The Math operation would continuously round up the computed value but leave the decimal on the range max. Using floor here ensures that we drop the decimal every time. This was my best attempt, though; I'm not sure it completely solves that problem but I haven't seen it appear again

miqa/core/tests/pyppeteer/test_scan_upload.py
annehaley: Changed in [0204ae3](https://github.com/OpenImaging/miqa/pull/563/commits/0204ae3d19d755b3fe8e7acbd71355c81b7a63be)

miqa/core/tests/pyppeteer/test_scan_upload.py
annehaley: Changed in [0204ae3](https://github.com/OpenImaging/miqa/pull/563/commits/0204ae3d19d755b3fe8e7acbd71355c81b7a63be)

web_client/src/components/ControlPanel.vue
```diff
@@ -206,52 +206,52 @@ export default {
...
+                class="d-flex px-5 py-3"
+                style="flex-direction: column; flex-grow: 3"
+              >
+                <div class="current-info-container">
+                  <div>
                     Project:
-                  </v-col>
-                  <v-col cols="9">
+                  </div>
+                  <div id="projectName">
```
zachmullen: Using `id` in a component is probably a code smell -- they are supposed to be globally unique in the DOM, so it makes this component essentially a singleton. I suggest we use a class instead.

zachmullen (changes requested):

web_client/src/components/ControlPanel.vue
annehaley: I see, okay. I thought it would make the XPath selection more granular, but I understand your point. I'll switch it back to a class selection

annehaley: Changed in [c155471](https://github.com/OpenImaging/miqa/pull/563/commits/c1554714d764c6a9d3f6b5665c6861b316096a5c)

web_client/src/components/ControlPanel.vue
annehaley: Changed in [ce6e59e](https://github.com/OpenImaging/miqa/pull/563/commits/ce6e59e61637edf4ce55cc41b430f93f70a9fc28)

web_client/src/components/ControlPanel.vue
annehaley: Changed in [efcdc53](https://github.com/OpenImaging/miqa/pull/563/commits/efcdc53c52a74e1f09983e7ecc08393b4cd31f67)

web_client/src/components/DataImportExport.vue
annehaley: Changed in [c2646a4](https://github.com/OpenImaging/miqa/pull/563/commits/c2646a4d439639257076c7cb52a46a2148540dc3)

web_client/src/components/ControlPanel.vue
zachmullen: It looks like be0b4a95cff91f9a596ed7261c40aa1e336f23b1 actually reverted the change from [efcdc53](https://github.com/OpenImaging/miqa/pull/563/commits/efcdc53c52a74e1f09983e7ecc08393b4cd31f67)

zachmullen (approved):
LGTM once the `align-center` is reintroduced

web_client/src/components/ExperimentsView.vue
annehaley: Changed in [ffed89e](https://github.com/OpenImaging/miqa/pull/563/commits/ffed89ed8303d5412e440f273887e9e308e2d928)

PR #567, opened 2022-08-23
Bug hunt 4

### LINKED ISSUE #427: Render CSV validation errors better
CSVs with newlines (`\n`) prepended/appended to the path column will report that the file cannot be found, which is fairly reasonable behavior. However, the error reported in the UI simply embeds the path directly in the HTML with no `<pre>` formatting, meaning that the HTML renderer eats the `\n`. This error is hard to debug without opening the CSV directly in a text editor.

We should either display the erroneous paths better, or simply trim whitespace from cells.

### LINKED ISSUE #513: Stale user browser tab authentication issue?
Lately when I have a stale MIQA front-end browser tab open, when I check it, I have been logged out.  That is good.  However, when I re-enter my credentials, I then receive a CSRF Forbidden error.  Refreshing the page didn't work for me (at least once), I believe. When I open a fresh tab instead, the login works. 
Screen Shot 2022-05-27 at 11 12 33 AM

### LINKED ISSUE #535: Automatic logout without warning even while using app

### LINKED ISSUE #555: Default window/level algorithm
Here is the code of how Slicer chooses default Window/Level:
https://github.com/Slicer/Slicer/blob/a5f75351073ef62fd6198d9480d86c0009d70f9b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L758-L790

Slicer constructs a histogram of intensities, and choose the min/max to correspond to 0.1% to 0.99% of pixels. If we don't want to deal with histograms, just taking 1% to 80% of intensity should be a reasonable approximation.

### LINKED ISSUE #565: Window/level controls only affecting one window
Environment: Chrome **with an Apple Trackpad** (the trackpad aspect could be important here)

Steps to reproduce:

1. Go to [a scan on the demo site](https://miqa.miqaweb.io/#/8de7d826-bd20-469c-b795-1dd7357d279d/8ab9ef76-37e3-4f55-85cb-0929d2f1feac)
2. Use the mouse drag motion in the Sagittal view to change the window / level. The first time, the other views correctly change.
3. Use the mouse drag motion a second time in the Sagittal view to change the window / level. 

Symptoms:

* After the first time the window / level is changed in the Sagittal view, the other views do not update correctly.
* The start of the drag with the trackpad also seems to be triggering a "click" event that is causing the crosshair to move and the other panels to change slices accordingly.

Let me know if this is hard to reproduce without a trackpad and we can work something out.

annehaley:
1. **experiments appearing twice**: I added a simple if statement to prevent experiments from getting added to the list twice; this appears to be a client side issue (I'm guessing these experiments and scans don't actually have duplicates on the server, since Aashish was able to get it to go away with a refresh). I think what was happening was a matter of a race condition as all the objects in the db are getting copied to the client state.

2. **setting window automatically**: I updated the logic to use the distribution of values in the image to calculate the middle and eliminate outliers from the auto range. The new auto range is +/- 2 sigma from the mean. Resolves #555.

3. **automatic sign off**: It seems that this was happening because the client was getting a 401 response from the server after 30 minutes and logging out from that. Our timer for the session timeout dialog was only paying attention to time elapsed between requests and client interactions, and it was ignorant of the server session timeout. I used this dialog to communicate to the user when their server session had expired, and allow them the option to continue (renew) the session or proceed to logout. Resolves #535 and resolves #513.

4. **scans marked as unreviewed**: I removed a bit of server logic (where we determine the state of a scan) that checked whether the existing latest decision has a creator. In the case of decisions brought in via import, they have no creator. That's why they were getting marked as if they had no decisions.

5. **project task overview not updating**: I updated the logic for fetching the state of a project from the server and invoke the `updateTaskOverview` method when a scan decision is submitted 

6. **emails sent to admins about new users are incomplete**: I reformatted the strings for the email content, which were getting cut off 

7. Resolves #565 : Removed drag interaction for updating window from vtk viewers

8. Resolves #427 : Strip whitespace from path values in an import file


web_client/src/components/WindowWidget.vue
```diff
@@ -66,8 +66,16 @@ export default defineComponent({
 
     function autoRange() {
       if (windowLocked.value) return;
-      currentRange.value = [widthMin.value, widthMax.value];
-      updateFromRange(currentRange.value);
+      const data = props.representation.getInputDataSet();
+      const distribution = data.computeHistogram(data.getBounds());
+      console.log(distribution);
```
zachmullen:
```suggestion
```

annehaley: Removed in [a840488](https://github.com/OpenImaging/miqa/pull/567/commits/a840488c70d450e1527858317c85212cc7333467)

miqa/core/models/project.py
```diff
@@ -102,20 +102,17 @@ def get_status(self):
...
-            .filter(last_decider_id__in=tier_2_reviewers)
-        )
+        complete_scans_in_project = [
+            scan
+            for (scan, last_decision) in [
+                (scan, ScanDecision.objects.filter(scan=scan).order_by('created')[0])
+                for scan in scans_in_project.filter(decision_count__gt=0)
+            ]
+            if last_decision.decision == 'U' or last_decision.creator.id in tier_2_reviewers
+        ]
```
zachmullen: We should do this as a single query rather than using one query per scan. I think we just need to modify the previous subquery.

zachmullen (changes requested):

miqa/core/models/project.py
annehaley: I'm having trouble writing this; I'm finding it difficult to filter on those two things with a single subquery. Could I have your help with the rewrite?

zachmullen: Sure thing. Let me take a crack at it, these things take some creativity to reason out.

miqa/core/models/project.py
```diff
@@ -97,22 +97,23 @@ def get_status(self):
...
+                ScanDecision.objects.filter(scan__id=models.OuterRef('id'))
+                .order_by('-created')
+                .values('creator_id')[:1]
+            ),
+            latest_decision=models.Subquery(
+                ScanDecision.objects.filter(scan__id=models.OuterRef('id'))
+                .order_by('-created')
+                .values('decision')[:1]
+            ),
+        ).filter(models.Q(latest_decider_id__in=tier_2_reviewers) | models.Q(latest_decision='U'))
```
zachmullen, line 112: This is a somewhat complicated query, but it emits the SQL that I was intending:

```sql
SELECT COUNT(*)
  FROM (
        SELECT "core_scan"."id" AS Col1
          FROM "core_scan"
         INNER JOIN "core_experiment"
            ON ("core_scan"."experiment_id" = "core_experiment"."id")
         WHERE ("core_experiment"."project_id" = '323e798a-3d2b-4d4e-9ab7-cde1206b5bbb'::uuid AND ((SELECT U0."creator_id" FROM "core_scandecision" U0 WHERE U0."scan_id" = "core_scan"."id" ORDER BY U0."created" DESC LIMIT 1) IN (1) OR (SELECT U0."decision" FROM "core_scandecision" U0 WHERE U0."scan_id" = "core_scan"."id" ORDER BY U0."created" DESC LIMIT 1) = 'U'))
       )
```

Note that I was also able to simplify the `scans_in_project` query to get rid of the annotations that required aggregation.

Because this is a complex (perhaps tending towards inscrutable) query, we should definitely write a unit test for this `get_status` method. The test should use our factories to create a Project with several Scans and ScanDecisions, and assert that the `get_status` counts come back correctly. I leave the creation of that test to you @annehaley 
😄

annehaley: Test added in [d4a8a0a](https://github.com/OpenImaging/miqa/pull/567/commits/d4a8a0a5c07c29a1e00ba6d9998d090fd6e97524)

miqa/core/tests/test_rest.py
```diff
@@ -62,6 +62,34 @@ def test_projects_list(user_api_client, project, user):
...
+    ]
+    scans = [scan_factory(experiment=experiment) for i in range(len(decisions))]
+    for i, scan in enumerate(scans):
+        decider = user_factory()
+        assign_perm(decisions[i][1], decider, project)
+        scan_decision_factory(scan=scan, creator=decider, decision=decisions[i][0])
+    resp = user_api_client().get(f'/api/v1/projects/{project.id}')
+    if resp.status_code == 200:
+        assert resp.data['status']['total_scans'] == len(scans)
+        assert resp.data['status']['total_complete'] == 3
```
zachmullen: Rather than testing with the API client, let's test the `project.get_status()` method directly, since the value of this test is in validating the query logic.

miqa/core/tests/test_rest.py
```diff
@@ -62,6 +62,34 @@ def test_projects_list(user_api_client, project, user):
...
+    # 3 of 5 scans marked as complete
+    decisions = [
+        ('U', 'tier_1_reviewer'),
+        ('U', 'tier_2_reviewer'),
+        ('UN', 'tier_1_reviewer'),  # needs tier 2 review
+        ('UN', 'tier_2_reviewer'),
+        ('Q?', 'tier_1_reviewer'),  # needs tier 2 review
+    ]
+    scans = [scan_factory(experiment=experiment) for i in range(len(decisions))]
+    for i, scan in enumerate(scans):
```
zachmullen:
```suggestion
    for decision, scan in zip(decisions, scans):
```

miqa/core/tests/test_rest.py
```diff
@@ -62,6 +62,34 @@ def test_projects_list(user_api_client, project, user):
...
+        decider = user_factory()
+        assign_perm(decisions[i][1], decider, project)
+        scan_decision_factory(scan=scan, creator=decider, decision=decisions[i][0])
+    resp = user_api_client().get(f'/api/v1/projects/{project.id}')
+    if resp.status_code == 200:
```
zachmullen: It's usually not good to have conditional blocks inside the test body. In this case, wouldn't we just want to assert that the status code was 200, rather than silently passing if the request fails?

zachmullen (changes requested):

miqa/core/tests/test_rest.py
annehaley: Changed in [491a19d](https://github.com/OpenImaging/miqa/pull/567/commits/491a19da608cc75902e53ae9f80bf0327f45144b)

miqa/core/tests/test_rest.py
annehaley: Changed in [491a19d](https://github.com/OpenImaging/miqa/pull/567/commits/491a19da608cc75902e53ae9f80bf0327f45144b)

miqa/core/tests/test_rest.py
annehaley: I agree. I had put this there because our `user_api_client` fixture is parameterized, so the test will run once for a superuser-authenticated client, once for a tier_2_reviewer-authenticated, once for tier_1_reviewer-authenticated, once for collaborator-authenticated, and once for no project access. In the last case, the project endpoint returns a `404`.
But, with the other change you suggested, we don't need to check the status code because we don't access the endpoint at all. And without using the `user_api_client` fixture, the test only runs once :)

miqa/core/tests/test_rest.py
```diff
@@ -62,6 +62,33 @@ def test_projects_list(user_api_client, project, user):
...
+        ('UN', 'tier_1_reviewer'),  # needs tier 2 review
+        ('UN', 'tier_2_reviewer'),
+        ('Q?', 'tier_1_reviewer'),  # needs tier 2 review
+    ]
+    scans = [scan_factory(experiment=experiment) for i in range(len(decisions))]
+    for decision, scan in zip(decisions, scans):
+        decider = user_factory()
+        assign_perm(decision[1], decider, project)
+        scan_decision_factory(scan=scan, creator=decider, decision=decision[0])
+    project.refresh_from_db()
```
zachmullen:
```suggestion
```

I think this line might not be necessary

annehaley: You're right. I had figured the creation of subsequent objects may need a refresh, but it works without it, so I suppose that's only necessary for changes directly to the object. Changed in [ed33350](https://github.com/OpenImaging/miqa/pull/567/commits/ed33350fc09eb7a9b0d3c12d0434e45943cfb627)

zachmullen (approved):

PR #573, opened 2022-08-26
Pre-release fixes

### LINKED ISSUE #569: Text overflow on a narrow window
Screenshot 2022-08-25 14 29 42
In the screenshot above, text overflows to the second line (red arrows) despite room in the same line (green).

The slider is hard to slide due to endpoint mouse capture range, but I guess this is hard to fix.

### LINKED ISSUE #571: Import or export should implicitly save settings
Changing paths on the project's settings, then clicking import or export should be equivalent to first clicking save, then clicking import/export. When a path is changed, but save is not clicked, the old paths are used and just-entered paths are reverted to the old content, effectively ignoring user input.

annehaley:
Resolves #569 
Resolves #571

### DIFF
--- a/web_client/src/components/ControlPanel.vue
+++ b/web_client/src/components/ControlPanel.vue
@@ -323,81 +323,80 @@ export default {
                     fill-height
                     fluid
                   >
-                    <v-row no-gutters>
-                      <v-col cols="2">
-                        Scan:
-                      </v-col>
-                      <v-col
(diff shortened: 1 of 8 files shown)


zachmullen (approved):
It would be better (long term) to use the Vue event bus system to manage the connection between request failures and the snackbar, but that would be a significant refactor, and I don't want it to hold this branch up.

PR #574, opened 2022-08-26
Make Artifacts Dynamic on Per Project Basis

davidshq:
Artifacts were previously set in code. This PR makes Artifacts dynamically selectable on a per project basis. 

(This dynamism is limited to the main application, it does not extend to the learning/neural network portion)

This was accomplished by adding two models to MIQA in Django.

- Artifact - represents individual artifacts
- Group - a general group model currently used only to group artifacts but with the potential to be used elsewhere (e.g. `default_evaluation_model_mapping` in `project.py` and `SCAN_TYPES` in `scan.py`)

Each Project can be associated with a Group (of artifacts). This group of artifacts is then displayed within the Vue UI (for selection by QA) and also utilized during `tasks.py` import.


miqa/core/models/group.py
```diff
@@ -0,0 +1,17 @@
+from uuid import uuid4
+from django.db import models
+
+
+class Group(models.Model):
```
zachmullen: Let's name this model `ArtifactGroup`

miqa/core/models/group.py
```diff
@@ -0,0 +1,17 @@
...
+    """Models a generic group of items
+
+        Allows one to reference a specific grouping of items.
+
+        It is currently used by Project to select a collection of Artifacts.
+    """
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
+    name = models.CharField(max_length=255, blank=False)
+    description = models.TextField()
```
zachmullen:
```suggestion
    name = models.CharField(max_length=255)
    description = models.TextField(blank=True)
```

`blank` defaults to `False`. (I assumed you meant to allow the description to be blank, but maybe not?)

miqa/core/models/project.py
```diff
@@ -54,6 +54,7 @@ class Project(TimeStampedModel, models.Model):
     )
     evaluation_models = models.JSONField(default=default_evaluation_model_mapping)
     default_email_recipients = models.TextField(blank=True)
+    artifact_group = models.ForeignKey('Group', null=True, on_delete=models.SET_NULL)
```
zachmullen: I assume the rationale for having a separate grouping relation is that you want to be able to have multiple projects use the same artifact group, is that correct?

miqa/core/rest/project.py
```diff
@@ -47,6 +52,18 @@ def get_permissions(self, obj):
 
         return permissions
 
+    def get_artifacts(self, obj):
+        if obj.artifact_group_id is not None:
+            artifacts = Artifact.objects.filter(group__artifact__group__id=obj.artifact_group_id)
```
zachmullen:
```suggestion
            artifacts = obj.artifact_group.artifacts
```

(That assumes you've set a `related_name` of `'artifacts'` on the ForeignKey from Artifact to Group.)

miqa/core/models/scan_decision.py
```diff
@@ -37,8 +26,13 @@ class ArtifactState(Enum):
     ABSENT = 0
     UNDEFINED = -1
 
+def default_identified_artifacts(scan_project_artifacts = ''):
```
zachmullen: I think rather than use this as the default value of `user_identified_artifacts`, we should redesign this a bit. Namely, let's make that column nullable, and if it's null, we can construct this value on the fly when we want to render it.

We can add the following to the `Project` model:

```python
@property
def artifacts(self) -> list[str]:
    if self.artifact_group:
        return [artifact.name for artifact in self.artifact_group.artifacts]
    else:
        return settings.DEFAULT_ARTIFACTS
```

and move the rest of the logic into the rendering layers (either the REST API or the front-end) since it's just string munging for display purposes.

miqa/core/rest/other_endpoints.py
```diff
@@ -27,7 +27,7 @@ class MIQAConfigView(APIView):
     def get(self, request):
         return Response(
             {
-                'artifact_options': [key for key in default_identified_artifacts()],
+                'artifact_options': list(default_identified_artifacts()),
```
zachmullen:
```suggestion
```

Let's remove this field, so that this is always read from the properties of the project.

miqa/core/rest/other_endpoints.py
```diff
@@ -27,7 +27,7 @@ class MIQAConfigView(APIView):
...
                 'auto_artifact_threshold': 0.4,
```
zachmullen: Should this threshold also be configurable on a per-project level, or are we OK with it remaining global?

zachmullen (changes requested):

miqa/core/models/artifact.py
```diff
@@ -0,0 +1,17 @@
...
+class Artifact(models.Model):
+    """Models an individual artifact
+
+        Artifacts are either (1) a visible alteration to an image (not part of the object imaged)
+        or (2) an unexpected deviation from the normative appearance of an object imaged.
+    """
+
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
+    name = models.CharField(max_length=255, blank=False)
+    group = models.ForeignKey('Group', on_delete=models.CASCADE)
```
zachmullen:
```suggestion
    group = models.ForeignKey('Group', on_delete=models.CASCADE, related_name='artifacts')
```

I tend to like this convention for related_names, rather than the default

miqa/core/models/project.py
davidshq: Correct, thanks.

miqa/core/models/group.py
davidshq: Correct, thanks.

miqa/core/models/project.py
davidshq: Yes, that is correct.

miqa/core/models/group.py
davidshq: @zachmullen - I had initially named this `ArtifactGroup` but realized there were multiple items that could use similar grouping. Rather than creating a separate model for each type of item I created the generic `Group`. 

Other objects I thought could be similarly grouped (/made dynamically editable) in the future include:
- evaluation models (core\project.py, learning/evaluation_models.py)
- scan types (core\scan.py)
- decision choices (core\scan_decision.py)
- columns (learning\nn_training.py)

If you don't see any of these as likely candidates to become dynamic in the future / would rather have a separate model for each, I can definitely refactor back to `ArtifactGroup`, just lmk.

zachmullen: I think a separate model for each would be preferable, otherwise it would be possible to group multiple different domain objects within the same Group.

dzenanz: Having those things dynamically editable sounds appealing to me.

curtislisle: @zachmullen , are you recommending that we shouldn't use the Group model and instead be specific to only artifacts, then create other groups for other changeable features?  I like that a project can have a particular Group name, which can specify Artifcacts now and later specify the other options Dave mentioned (scans, learning columns, etc.)  Maybe I am just confused, but I liked the design as originally submitted.  Maybe I just need a bit of explanation.  thanks.

zachmullen: Yes, that is what I was proposing. If we actually want multiple different types of things to be grouped together, then perhaps a generic `Group` is a good model, but that didn't seem like the design intent, as the field on the Project model was named `artifact_group`. I assumed we would only ever want to group things together of the same domain type.

curtislisle: @zachmullen , Thanks.  It is my opinion that we would want to use the Group model to select sets of different things to go with projects.  In other words, the same way we have an Artifact model now with a group attribute, so we can build lists of Artifcats for each project , I think we might want to have the name of the model and its Evaluation columns (the outputs of the learning model) done similarly.    

I don't want to embark on this approach immediately across the board, but I'd like to leave room for generalizing the application of MIQA to a learning model designed for my Lung Cancer application, for example.

zachmullen: If we do want multiple types of domain objects to be put into the same group, I'd still suggest we change the name to e.g.  `ProjectOptionsGroup`, to avoid collision with the existing `Group` model from django-contrib.

davidshq: @zachmullen - Definitely see how the `artifact_group` on Project could be misleading. If I understand correctly Django can support multiple fields to the same model with different names (e.g. `scan_group` also pointing to the `Group` model).

Changing the name from `Group` makes sense to me. Wondering if we should make it `ProjectOptionsGroup` or `OptionsGroup` to allow it to be used more generically for settings that could occur at other levels (e.g. global or perhaps per experiment)? I don't have a strong opinion on this, just wanted to raise the Q.

davidshq:
FYI Update: Ran into some issues and needed to rebuild my environment. Getting close and should have some commits to finish request changes and clean things up coming this week.

miqa/core/rest/other_endpoints.py
davidshq: This makes sense to me. Maybe a separate issue can be added? I'd like to avoid expanding the scope atm.

zachmullen: Definitely, let's do it on a follow-up.

miqa/core/models/group.py
davidshq: Group has been recreated as SettingsGroup (meant it to be OptionsGroup but if it is all the same will leave as is, otherwise, easy change to OptionsGroup).

miqa/core/rest/other_endpoints.py
davidshq: I ran into some weird issues implementing this on the Vue side, so I've reverted to using `default_identified_artifacts` for now. I'd love to come back to it at some point, but need to keep moving on lung's project.  Understand if you don't want to merge without this implemented.

miqa/core/models/scan_decision.py
davidshq: Removing default_identified_artifacts and it's string munging is outside what I could accomplish atm. Totally good if you don't want to merge without, as mentioned elsewhere have to move onto other lung project priorities.

curtislisle:
Thanks. Agreed.

davidshq:
I completely rewrote the dynamic settings functionality from the ground up, so I'm going to close this PR.

The rewritten functionality can be found in: https://github.com/davidshq/miqa/tree/abstract-with-enums

If there is interest in merging it, let me know and I can create a new PR.

The new functionality uses a single model (Setting) that can support almost all types of settings.

There is a `setting.json` file in the root of this branch which can be used with `./manage.py loaddata` to add all the settings in without needing to do so manually.

I've focused on artifacts, models, file type to model mappings, etc.

PR #583, opened 2022-09-14
Add json case to export method

### LINKED ISSUE #581: Export does not save file in the JSON format
Report by @kipohl against v1.0.5

annehaley:
Resolves #581 

I actually didn't account for the JSON case in the export method at all; it was saving with the CSV format no matter what the file extension was. It was therefore a quick fix to get it working.

### DIFF
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -318,6 +318,7 @@ def perform_export(project_id: Optional[str]):
 
     if project_id is None:
         # A global export should export all projects
+        project = None
         projects = list(Project.objects.all())
         export_path = GlobalSettings.load().export_path
     else:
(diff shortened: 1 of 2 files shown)


miqa/core/tasks.py
```diff
@@ -376,8 +377,17 @@ def perform_export(project_id: Optional[str]):
...
                 )
     export_df = pandas.DataFrame(data, columns=IMPORT_CSV_COLUMNS)
+
     try:
-        export_df.to_csv(export_path, index=False)
+        if export_path.endswith('csv'):
+            export_df.to_csv(export_path, index=False)
+        elif export_path.endswith('json'):
+            json_contents = import_dataframe_to_dict(export_df, project)
+            json.dump(json_contents, open(export_path, 'w'))
```
zachmullen:
```suggestion
            with open(export_path, 'w') as fd:
                json.dump(json_contents, fd)
```

That will explicitly ensure that we don't leak file descriptors.

zachmullen (changes requested):

zachmullen (approved):

PR #591, opened 2022-09-16
Use setting to determine if null dates should be replaced

### LINKED ISSUE #590: Setting to replace null dates on import should be disabled by default
From @kipohl via Slack:
> "created" was Null and then after it is exported it has a date in it

### LINKED ISSUE #592: Reimporting project file exported by MIQA very slightly changes the timestamp
**How to reproduce?**
The most problematic issue is that if you export a file from miqa, log out of miqa, log in and import the file (that you exported), and then export it again, it again produces a different file.

annehaley:
Resolves #590, resolves #592


zachmullen (approved):

miqa/core/tasks.py
```diff
@@ -49,6 +48,19 @@ def _download_from_s3(path: str, public: bool) -> bytes:
     return buf.getvalue()
 
 
+def datetime_string_valid(text):
+    date_formats = ['%Y-%m-%d', '%d/%m/%Y']
+    time_formats = ['%H:%M', '%H:%M:%S', '%H:%M:%S%f', '%H:%M:%S%f%z']
+    for df in date_formats:
+        for tf in time_formats:
+            try:
+                print(f'{df} {tf}')
```
zachmullen:
```suggestion
```

miqa/core/tasks.py
```diff
@@ -49,6 +48,19 @@ def _download_from_s3(path: str, public: bool) -> bytes:
     return buf.getvalue()
 
 
+def datetime_string_valid(text):
```
zachmullen: Can we just pull in the [dateparser](https://pypi.org/project/dateparser/) library and throw the string in, and if it can't figure it out, then we declare the date invalid? That seems easier than trying to enumerate all the valid date types, unless of course we want to restrict it to just these.

miqa/core/tasks.py
```diff
@@ -213,19 +225,17 @@ def perform_import(import_dict):
...
-                            try:
-                                datetime.datetime.strptime(
-                                    last_decision_dict['created'], '%Y-%m-%d'
-                                )
-                            except ValueError:
-                                created = timezone.now()
-                            else:
-                                created = last_decision_dict['created']
+                            valid_dt = datetime_string_valid(last_decision_dict['created'])
+                            print(valid_dt)
```
zachmullen:
```suggestion
```

zachmullen (changes requested):

miqa/core/tasks.py
annehaley: Ooh, cool library. I haven't used it before. I'll try it out.

zachmullen (approved):

PR #594, opened 2022-09-21
Remove unused advanceLoop method on Frame

davidshq:
Code cleanup, removes an orphaned function `advanceLoop`. It isn't called elsewhere and it attempts to call two non-existent functions (`updateFrame` and `nextAnimRequest`).


annehaley (approved):

annehaley:
The source branch is behind; current master does not contain the `advanceLoop` function. Closing this.

PR #597, opened 2022-09-26
Load the next scan each time a new scan is viewed

annehaley:
Each time a new scan is requested, we queue loading all the frames in that scan, as well as all the frames in the next scan. If the scan has been loaded already, we skip the queue of that scan.
Finding the next scan to queue includes the first scan in the next experiment if the current scan is the last in its experiment. 

From @zachmullen via Slack 
> We basically want a fixed-sized ring buffer of prefetched scans that goes forward to future experiments

How explicitly should we define the ring buffer? We previously didn't worry about clearing cached frames out as we moved along, perhaps we can trust the browser to manage the cleanup. Within this PR, I only implemented the _addition_ of items to the buffer. Do we need to manually remove items?

I'd appreciate extra scrutiny on your review; I'm sure there's much to improve.

### DIFF
--- a/web_client/src/store/index.ts
+++ b/web_client/src/store/index.ts
@@ -29,6 +29,7 @@ Vue.use(Vuex);
 const fileCache = new Map();
 const frameCache = new Map();
 let readDataQueue = [];
+const loadedData = [];
 const pendingFrameDownloads = new Set<any>();
 const poolSize = Math.floor(navigator.hardwareConcurrency / 2) || 2;
 let taskRunId = -1;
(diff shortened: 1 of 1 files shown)


zachmullen:
We'll definitely want to evict old frame data out of the cache. IMO we can do that as soon as they move on to the next scan, or maybe we can keep one previous scan around. If we never evict them, we'll run up against memory limits.

I haven't explored the code much; where is the cache data structure located?

curtislisle:
Maybe we can make the cache configurable at the Django settings level. For the annotation application, I expect to be moving back and forth between scans within an experiment so would like to keep all scans hot,  but keeping  a max of X (X=3? configurable?) experiments could allow us to flush while still minimizing wait time for users.

zachmullen:
Thanks for the feedback @curtislisle . I didn't realize going backward and forward would be common -- in that case, a ring buffer makes even more sense as our cache structure.

curtislisle:
If it isn't too much trouble to implement a ring buffer, I like the
flexibility this brings.  However, I don't think it is a critical need,
though, so we can prioritize accordingly.  The Stanford application will be
forward only direction.  I am planning for the annotation application to
examine multiple scans within an experiment simultaneously and may go
backwards occasionally.

annehaley:
> I haven't explored the code much; where is the cache data structure located?

The caching structure is complicated, but I think all functional parts are located in `store/index.js`. We do have map structures for `fileCache` and `frameCache` (unsure of the distinction and need for separation...). However we only clear them in the `reset` action, and otherwise do not reduce their size.
[reset in index.js](https://github.com/OpenImaging/miqa/blob/a5e57c6137bf148ad60ee5fea91417f5ff5247ee/web_client/src/store/index.ts#L590)

zachmullen:
I think this merits adding a cache abstraction, i.e. functions to mediate adding to the cache that internally handle eviction, rather than direct access to the underlying data structure.

annehaley:
I think this is a good idea, but I'm not sure how to begin. Do you have a similar example or would you mind working with me on this?

annehaley:
From @zachmullen  via Slack

> unfortunately I won't have time this week to work on a more robust cache / prefetch implementation with you. If what you have solves the immediate issue at hand, I would suggest merging it.

Merging this and opening #598 to track further action

PR #603, opened 2022-09-29
Put a spinner in Experiments view if experiment data is still loading

### LINKED ISSUE #600: Display "update in progress" instead of "no data"
When looking at a project, "no data" message is incorrectly displayed until the list is populated. Show a spinner here instead.

annehaley:
Resolves #600

### DIFF
--- a/web_client/src/components/ExperimentsView.vue
+++ b/web_client/src/components/ExperimentsView.vue
@@ -281,6 +281,13 @@ export default {
         </ul>
       </div>
       <div
+        v-else-if="currentProject.experiments.length"
+        class="pa-5"
+        style="width: 60%; text-align: center"
+      >
+        <v-progress-circular indeterminate />
+      </div>
+      <div
         v-else
         class="pa-5"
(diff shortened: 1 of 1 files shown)


web_client/src/components/ExperimentsView.vue
```diff
@@ -280,6 +280,13 @@ export default {
           </li>
         </ul>
       </div>
+      <div
+        v-else-if="currentProject.experiments.length"
```
zachmullen, line 284: Shouldn't we use an out-of-band `loading: boolean` state for this? What if the project simply has no experiments yet?

annehaley: If the project has no experiments, the `v-else-if` directive will evaluate to false and the `v-else` block will show "No imported data"

zachmullen: Oh, I see. Thanks for the explanation.

zachmullen (approved):

PR #605, opened 2022-09-29
Run import step synchronously

### LINKED ISSUE #593: After import, the project page does not update correctly
Reported by @kipohl 
https://files.slack.com/files-pri/T0AGVLRKQ-F042L1UT7HR/image.png

### LINKED ISSUE #601: Right after import, a project can show 0/0
After clicking reload button, it changes to correct count (e.g. 0/2).

annehaley:
Intended to resolve #593 and resolve #601

I was not able to replicate the problem described in these two issues, so this was my best attempt without being able to verify. 

My guess is that previously, the only synchronous part of the import was to verify the validity of the import file. Objects made by the import were done asynchronously. Thus, when the synchronous part was done, it returned to the client and the client immediately updated the task overview. If the asynchronous part had not finished making the objects, the task overview would show up as having no objects within the project.

With this change, the part of the import that creates objects is now done synchronously, so the client won't get an answer (and won't request the task overview) until the objects are done saving to the database.

The most time-consuming part of the import is the evaluation step, and this part is still done asynchronously.

### DIFF
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -156,7 +156,7 @@ def import_data(project_id: Optional[str]):
         raise APIException(f'MIQA lacks permission to read {import_path}.')
 
     import_dict, not_found_errors = validate_import_dict(import_dict, project)
-    perform_import.delay(import_dict)
+    perform_import(import_dict)
     return not_found_errors
 
 


miqa/core/tasks.py
```diff
@@ -156,7 +156,7 @@ def import_data(project_id: Optional[str]):
         raise APIException(f'MIQA lacks permission to read {import_path}.')
 
     import_dict, not_found_errors = validate_import_dict(import_dict, project)
-    perform_import.delay(import_dict)
+    perform_import(import_dict)
```
zachmullen, line 159: What's the rationale for this change? Is it possible import might take a while?

annehaley: Good question :grin: I was writing the above description just as you asked

zachmullen: Explanation makes sense. Thanks!

annehaley:
Not sure why https://github.com/OpenImaging/miqa/actions/runs/3153216022/jobs/5129423602 keeps failing...

zachmullen (approved):

PR #606, opened 2022-09-29
Sentry errors

annehaley:


web_client/src/components/ControlPanel.vue
```diff
@@ -411,7 +411,10 @@ export default {
                           :decision="decision"
                         />
                         <div
-                          v-if="currentViewData.scanDecisions.length === 0"
+                          v-if="
+                            currentViewData?.scanDecisions === undefined
+                              || currentViewData?.scanDecisions?.length === 0
```
zachmullen:
```suggestion
                            !currentViewData?.scanDecisions?.length
```

Is this equivalent?

annehaley: Yes, thanks for the short version!

annehaley: Actually, I'm discovering that neither works; we cannot use the `?` operator like this within the template. https://github.com/vuejs/vue/issues/11088.

zachmullen: If we did want to use that syntax, we could throw it into a computed property and use that in the template.

zachmullen (approved):

zachmullen:
LGTM. I think moving a lot of this to typescript would catch a lot of these errors at build time.

PR #610, opened 2022-10-07
Allow deletion of experiments within a project

annehaley:


miqa/core/rest/experiment.py
```diff
@@ -62,13 +62,17 @@ def validate(self, data):
...
     filter_backends = [filters.DjangoFilterBackend]
-    permission_classes = [IsAuthenticated, UserHoldsExperimentLock]
     serializer_class = ExperimentSerializer
 
+    def get_permissions(self):
+        if self.action == 'destroy' or self.action == 'create':
+            permission_classes = [IsAuthenticated]
+        else:
+            permission_classes = [IsAuthenticated, UserHoldsExperimentLock]
+        return [permission() for permission in permission_classes]
```
zachmullen, line 74: Do these have to be instances of the class, or is the class itself OK?

annehaley: I had pasted in this code from the drf docs: https://www.django-rest-framework.org/api-guide/viewsets/#:~:text=def%20get_permissions(,in%20permission_classes%5D

I know that when you specify `permission_classes` it's supposed to be the class itself, but examples I've seen of `get_permissions` return a list of instances.

zachmullen: Ah cool, TIL.

PR #613, opened 2022-10-10
Scan Navigation

annehaley:
Previously, the url for the main view contained the frame id, and the main view was different for each frame.
In order to  speed up the reaction time of the frame slider, the main view is routed by scan id. Changing the frame is a matter of changing the state only, not the whole main view.

### DIFF
--- a/web_client/src/components/ControlPanel.vue
+++ b/web_client/src/components/ControlPanel.vue
@@ -92,6 +92,7 @@ export default {
     ...mapMutations([
       'setShowCrosshairs',
       'setStoreCrosshairs',
+      'setCurrentFrameId',
     ]),
     openScanLink() {
       window.open(this.currentViewData.scanLink, '_blank');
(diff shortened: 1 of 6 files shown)


zachmullen:
Are we OK with losing the ability to permalink to individual frames?

What was the performance issue associated with the slider and updating of the hash?

annehaley:
Yes, we lose the ability to link to a frame, but I think the user only cares about linking to a scan. The frames are to be considered in order anyway, in the context of the whole scan. 
Removing the frame's association with the current URL speeds up the slider because the whole page doesn't need to be rerendered, only the vtk viewers.

From @kipohl via Slack:
> slider of frame has a strange behavior - it shows a couple of frames and then jumps to the frame that I moved the dot to . The speed should be almost as  fast it is for moving through the slices in axial , saggital , coronal - the slide should behave as those slides - that is how it was for miqa

zachmullen:
Thanks for the explanation. LGTM.

PR #616, opened 2022-10-13
Add S3StorageMixin to DockerComposeProductionConfiguration

annehaley:


miqa/settings.py
```diff
@@ -107,12 +108,19 @@ class DockerComposeProductionConfiguration(
...
 ):
     """For the production deployment using docker-compose."""
 
     MIQA_URL_PREFIX = values.Value(environ=True, default='/')
 
+    # Configure connection to S3 bucket by setting the following environment variables:
+    # AWS_S3_REGION_NAME
+    # AWS_S3_ACCESS_KEY_ID
+    # AWS_S3_SECRET_ACCESS_KEY
+    # AWS_STORAGE_BUCKET_NAME
```
zachmullen: The S3StorageMixin actually has different names for the environment variables, I think to be more in line with AWS SDK conventions, as well as Django storage conventions:

```suggestion
    # AWS_DEFAULT_REGION
    # AWS_ACCESS_KEY_ID
    # AWS_SECRET_ACCESS_KEY
    # DJANGO_STORAGE_BUCKET_NAME
```

zachmullen (changes requested):

miqa/settings.py
annehaley: Oh I see, I assumed the env vars had the same name as the python vars. Thanks for catching that.

zachmullen (approved):

PR #622, opened 2022-10-27
Possible cause of hanging pages #1

annehaley:


web_client/src/store/index.ts
```diff
@@ -494,11 +494,6 @@ const {
       }
       if (state.currentProject && taskOverview.project_id === state.currentProject.id) {
         state.currentTaskOverview = taskOverview;
-        Object.values(store.state.scans).forEach((scan: Scan) => {
-          if (taskOverview.scan_states[scan.id] && taskOverview.scan_states[scan.id] !== 'unreviewed') {
-            store.dispatch.reloadScan(scan.id);
```
zachmullen, line 499: So there was a chance that calling `reloadScan` could eventually cause this `setTaskOverview` to run?

annehaley: We know that refreshing the page will cause `loadProject` to execute, which calls `setTaskOverview`. I was thinking that the unreviewed state of these scans would cause `reloadScan` to execute, possibly causing an infinite loop. But it appears this is not the cause, as the problem is still there.

annehaley: This is the initiator for the `taskOverview` request that hangs: 
image

PR #627, opened 2022-10-28
Remove infinite loop case

annehaley:


web_client/src/store/index.ts
```diff
@@ -231,13 +231,12 @@ function queueLoadScan(scan, loadNext = 0) {
         ];
       }
     } else {
-      let offset = 1;
+      let newIndex = scansInSameExperiment.indexOf(scan.id);
       while (!nextScan || !includeScan(nextScan.id)) {
+        if (newIndex >= scansInSameExperiment.length) break;
```
zachmullen: I think it'd be more readable to just combine this into the loop condition, i.e.

```typescript
while ((!nextScan || !includeScan(nextScan.id)) && newIndex < scansInExperiment.length) {

Comment (annehaley) [reply web_client/src/store/index.ts]: Changed in [dbd0ac5](https://github.com/OpenImaging/miqa/pull/627/commits/dbd0ac55bde23992de78bbeb686fea9661ca6217)

File: web_client/src/store/index.ts [hunk 2]
```diff
@@ -231,13 +231,14 @@ function queueLoadScan(scan, loadNext = 0) {
         ];
       }
     } else {
-      let offset = 1;
-      while (!nextScan || !includeScan(nextScan.id)) {
+      let newIndex = scansInSameExperiment.indexOf(scan.id);
```
zachmullen: Is it guaranteed that `scan.id` will be present in `scansInSameExperiment`? Just trying to consider what would happen if `newIndex` became `-1` here, or if that's even possible.

annehaley: Hmm, i think it's guaranteed by the definition of `scansInSameExperiment`, but it couldn't hurt to protect ourselves. Changed in [5b9e1e3](https://github.com/OpenImaging/miqa/pull/627/commits/5b9e1e35f3a6889740f80b158c012c8c4a056864)

web_client/src/store/index.ts
```diff
@@ -231,10 +231,11 @@ function queueLoadScan(scan, loadNext = 0) {
         ];
       }
     } else {
-      let newIndex = scansInSameExperiment.indexOf(scan.id);
+      let newIndex = scansInSameExperiment.indexOf(scan.id) + 1;
```
zachmullen, line 234: I assume the `+ 1` here is because we want the index after the current scan?

annehaley: Yes, It should have gone in the last commit.

PR #630, opened 2022-10-31
Sentry and other bugs

annehaley:


zachmullen (approved):

web_client/src/components/Navbar.vue
```diff
@@ -72,7 +72,11 @@ export default defineComponent({
         <template #activator="{ on }">
           <span v-on="on">MIQA</span>
         </template>
-        <span>{{ MIQAConfig.version.length > 0 ? MIQAConfig.version : "Demo Instance" }}</span>
+        <span>{{
+          MIQAConfig.version && MIQAConfig.version.length > 0
```
zachmullen: Does this come from the server? Can we just build it into the client so that it's guaranteed to be string-valued?

annehaley: Perhaps like in [4709ace](https://github.com/OpenImaging/miqa/pull/630/commits/4709acebfd344b147405461225c923e2c2b9d0c5)? Is this what you mean?

web_client/src/components/Navbar.vue
```diff
@@ -72,10 +72,9 @@ export default defineComponent({
         <template #activator="{ on }">
           <span v-on="on">MIQA</span>
         </template>
-        <span>{{
-          MIQAConfig.version && MIQAConfig.version.length > 0
-            ? MIQAConfig.version
-            : "Demo Instance"
+        <span>{{ MIQAConfig.version.length > 0
+          ? MIQAConfig.version
+          : "Demo Instance"
```
zachmullen: Yep, this is what I meant, though I would then just change this to:

```
<span>{{ MIQAConfig.version || "Demo Instance" }}</span>

annehaley: Changed in [536e8b1](https://github.com/OpenImaging/miqa/pull/630/commits/536e8b1799d940291940589960350e4befa96ed6)

PR #633, opened 2022-11-11
Make swapToFrame method faster

annehaley:
This PR makes the frame slider respond faster to change.

### DIFF
--- a/web_client/src/store/index.ts
+++ b/web_client/src/store/index.ts
@@ -814,7 +814,6 @@ const {
         }
         const start = Date.now();
         sourceProxy.setInputData(frameData);
-        console.log('diff after set', Date.now() - start);
         if (needPrep || !state.proxyManager.getViews().length) {
           prepareProxyManager(state.proxyManager);
           state.vtkViews = state.proxyManager.getViews();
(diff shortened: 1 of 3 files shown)


web_client/src/store/index.ts
```diff
@@ -810,7 +812,9 @@ const {
           );
           frameData = result.frameData;
         }
+        const start = Date.now();
         sourceProxy.setInputData(frameData);
+        console.log('diff after set', Date.now() - start);
```
zachmullen:
```suggestion
```

zachmullen (changes requested):

web_client/src/store/index.ts
annehaley: Thanks!

zachmullen (approved):

web_client/src/store/index.ts
```diff
@@ -49,9 +50,9 @@ function prepareProxyManager(proxyManager) {
       view.setOrientationAxesVisibility(false);
       view.getRepresentations().forEach((representation) => {
         representation.setInterpolationType(InterpolationType.NEAREST);
-        representation.onModified(() => {
+        representation.onModified(macro.debounce(() => {
           view.render(true);
-        });
+        }, 0));
```
zachmullen, line 55: I think having a debounce with a 0 time parameter merits a comment explaining it.

annehaley: Does the comment in [b5444a7](https://github.com/OpenImaging/miqa/pull/633/commits/b5444a7cce695b048450122c2fde71fe8862021f) suffice? Or should I explain further about the simultaneous changes?

zachmullen: LGTM

PR #634, opened 2022-11-14
Fix mypy test failure

annehaley:


miqa/core/rest/frame.py
```diff
@@ -125,13 +125,21 @@ def create(self, request, *args, **kwargs):
             scan.save()
         elif 'scan' in serializer.data:
             scan = Scan.objects.get(id=serializer.data['scan'])
+            if not scan:
+                raise APIException(
+                    'Could not create new Frame.'
```
zachmullen: This should probably have status=404

miqa/core/rest/frame.py
```diff
@@ -125,13 +125,21 @@ def create(self, request, *args, **kwargs):
...
             if not get_perms(request.user, scan.experiment.project):
                 Response(status=status.HTTP_403_FORBIDDEN)
         else:
             raise APIException(
                 'Provide either a parent scan or grandparent experiment for this frame.'
             )
 
+        if not scan:
+            raise APIException(
+                'Could not create new Frame.'
```
zachmullen: Same here

miqa/core/rest/frame.py
zachmullen: Also, we should probably make the message more informative, like "Scan not found"

miqa/learning/nn_training.py
```diff
@@ -588,7 +588,7 @@ def train_and_save_model(df, count_train, save_path, num_epochs, val_interval, o
         logger.info('-' * 25)
         logger.info(f'epoch {epoch + 1}/{num_epochs}')
         model.train()
-        epoch_loss = 0
+        epoch_loss = float(0)
```
zachmullen: Another way to write this is as a literal, `0.0`, which would avoid the cast

annehaley: Changed in [107d673](https://github.com/OpenImaging/miqa/pull/634/commits/107d673714769932d133942ea8355def62151cc7)

miqa/core/rest/frame.py
annehaley: Changed in [1d6a5b5](https://github.com/OpenImaging/miqa/pull/634/commits/1d6a5b53988507eca80fd0184717ac0fbd3534e4)

miqa/core/rest/frame.py
annehaley: Changed in [1d6a5b5](https://github.com/OpenImaging/miqa/pull/634/commits/1d6a5b53988507eca80fd0184717ac0fbd3534e4)

zachmullen (approved):

PR #635, opened 2022-11-15
Add automatic reset demo task

annehaley:
This only applies to instances where `DEMO_MODE=True`

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: ./dev/django.Dockerfile
-    command: ["./manage.py", "runserver", "0.0.0.0:8000"]
+    command: [ "./manage.py", "runserver", "0.0.0.0:8000" ]
     # Log printing via Rich is enhanced by a TTY
     tty: true
     env_file: ./dev/.env.docker-compose
@@ -30,13 +30,17 @@ services:
     build:
       context: .
       dockerfile: ./dev/django.Dockerfile
-    command: [
-      "celery",
-      "--app", "miqa.celery",
-      "worker",
-      "--loglevel", "INFO",
-      "--without-heartbeat"
-    ]
+    command:
+      [
+        "celery",
+        "-A",
+        "miqa.celery",
+        "worker",
+        "--loglevel",
+        "INFO",
+        "--without-heartbeat",
+        "-B" # add celery beat worker to schedule automatic tasks
+      ]
     # Docker Compose does not set the TTY width, which causes Celery errors
     tty: false
     env_file: ./dev/.env.docker-compose
@@ -50,7 +54,7 @@ services:
 
   npm:
     image: node:latest
-    command: ["npm", "run", "serve"]
+    command: [ "npm", "run", "serve" ]
     working_dir: /opt/client
     environment:
       # Fixes https://github.com/OpenImaging/miqa/issues/298
--- a/miqa/celery.py
+++ b/miqa/celery.py
@@ -1,7 +1,9 @@
+from datetime import timedelta
 import os
 
 from celery import Celery
 import configurations.importer
+from django.conf import settings
 
 os.environ['DJANGO_SETTINGS_MODULE'] = 'miqa.settings'
 if not os.environ.get('DJANGO_CONFIGURATION'):
@@ -14,3 +16,11 @@ app = Celery(config_source='django.conf:settings', namespace='CELERY')
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+if settings.DEMO_MODE:
+    app.conf.beat_schedule = {
+        'reset-demo': {
+            'task': 'miqa.core.tasks.reset_demo',
+            'schedule': timedelta(days=1),
+        }
+    }
--- a/miqa/celery.py
+++ b/miqa/celery.py
@@ -1,9 +1,7 @@
-from datetime import timedelta
 import os
 
 from celery import Celery
 import configurations.importer
-from django.conf import settings
 
 os.environ['DJANGO_SETTINGS_MODULE'] = 'miqa.settings'
 if not os.environ.get('DJANGO_CONFIGURATION'):
@@ -16,11 +14,3 @@ app = Celery(config_source='django.conf:settings', namespace='CELERY')
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
-
-if settings.DEMO_MODE:
-    app.conf.beat_schedule = {
-        'reset-demo': {
-            'task': 'miqa.core.tasks.reset_demo',
-            'schedule': timedelta(days=1),
-        }
-    }
--- a/miqa/core/tasks.py
+++ b/miqa/core/tasks.py
@@ -50,6 +50,21 @@ def _download_from_s3(path: str, public: bool) -> bytes:
 
 
 @shared_task
+def reset_demo():
+    if not settings.DEMO_MODE:
+        raise Exception('Demo Mode is false; noop.')
+    try:
+        demo_project = Project.objects.get(name='Demo Project')
+    except Project.DoesNotExist:
+        raise Exception('No Demo Project; noop.')
+    demo_project.import_path = 's3://miqa-storage/miqa.csv'
+    demo_project.export_path = 'samples/demo.json'
+    demo_project.save()
+    import_data(demo_project.id)
+    Project.objects.exclude(id=demo_project.id).delete()
+
+
+@shared_task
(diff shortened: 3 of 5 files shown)


miqa/core/tasks.py
```diff
@@ -49,6 +49,21 @@ def _download_from_s3(path: str, public: bool) -> bytes:
     return buf.getvalue()
 
 
+@shared_task
+def reset_demo():
+    if not settings.DEMO_MODE:
+        raise Exception('Demo Mode is false; noop.')
```
brianhelba:
```suggestion
```

I'd suggest removing this sort of defensive programming wherever possible. You should be able to trust that internal APIs are called by other parts of the program with the correct preconditions. To verify this, you can add tests using things like [`Mock.assert_called_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with), which [pytest-mock](https://pytest-mock.readthedocs.io/) can help to set up.

I realize that this case is more complicated, since whether this function gets called depends entirely on a runtime configuration which necessarily could be different in test vs prod. However, do you not trust your own present configuration (in which case, I think we could devise a test) or do you not trust some future developer (maybe future you) to continue to configure and invoke this directly?

Of course, all these are trade-offs, I'm just trying to make sure defensive lines of code are really, really needed, because otherwise they tend to grow out of control.

brianhelba:
```suggestion
    if not settings.DEMO_MODE:
        raise RuntimeError('Demo Mode is false; noop.')
```

If you want to keep this, consider [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError), which is a slightly more specific type.

miqa/core/tasks.py
```diff
@@ -49,6 +49,21 @@ def _download_from_s3(path: str, public: bool) -> bytes:
...
+    try:
+        demo_project = Project.objects.get(name='Demo Project')
+    except Project.DoesNotExist:
+        raise Exception('No Demo Project; noop.')
```
brianhelba:
```suggestion
    demo_project = Project.objects.get(name='Demo Project')
```

I'm not clear on the value of transforming the exception type and message. This exception is for the benefit of other developers and `Project.DoesNotExist`, raised from this point in the code or just within the `reset_demo` function, seems pretty obvious already. Maybe add a code comment if you really don't think it's apparent why the code expects that exactly one demo Project should exist.

brianhelba:
```suggestion
    except Project.DoesNotExist as e:
        raise Exception('No Demo Project; noop.') from e
```

If you want to keep this, [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) is a nice way to preserve the original exception too (Sentry will show both).

miqa/celery.py
```diff
@@ -14,3 +16,11 @@
...
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+if settings.DEMO_MODE:
+    app.conf.beat_schedule = {
+        'reset-demo': {
+            'task': 'miqa.core.tasks.reset_demo',
+            'schedule': timedelta(days=1),
+        }
+    }
```
brianhelba: By placing this here, you're making this code execute at import time. Whenever possible, it'd be nice to avoid import side effects, though I realize Celery and Django both somewhat abuse this.

Also, I'd suggest using the Django settings registry as the source of truth for all Celery config, since that's exactly what
```python
config_source='django.conf:settings', namespace='CELERY'
```
is setting up. It also makes it easier to test and debug settings if you know they're all guaranteed to be defined at the same place. You'd just have to use the Django setting named as `CELERY_BEAT_SCHEDULE`.

If you want the value of this setting to dynamically depend on another setting, [Django Configurations allows the use of `@property`](https://django-configurations.readthedocs.io/en/stable/patterns.html#property-settings). You could do:
```python
class MiqaMixin(ConfigMixin):
    ...

    @property
    def CELERY_BEAT_SCHEDULE(self):
        if self.DEMO_MODE:
            return {
                'reset-demo': {
                    'task': 'miqa.core.tasks.reset_demo',
                    'schedule': timedelta(days=1),
                }
            }
        else:
            return {}
```

Alternatively, you could have something like:
```python
class MiqaMixin(ConfigMixin):
    ...

    CELERY_BEAT_SCHEDULE = {}

    @staticmethod
    def before_binding(configuration):
        if configuration.DEMO_MODE:
            configuration.CELERY_BEAT_SCHEDULE.update({
                'reset-demo': {
                    'task': 'miqa.core.tasks.reset_demo',
                    'schedule': timedelta(days=1),
                }
            })
```

which is a little more complicated, but more extensible.

docker-compose.override.yml
```diff
@@ -30,13 +30,17 @@ services:
...
+    command:
+      [
+        "celery",
+        "-A",
+        "miqa.celery",
+        "worker",
+        "--loglevel",
+        "INFO",
+        "--without-heartbeat",
+        "-B" # add celery beat worker to schedule automatic tasks
```
brianhelba, line 42:
```suggestion
```

Rather than always running the `celery` container with `-B`, I'd suggest adding a totally new container (named something like `celerybeat`) with the exactly same other options as the `celery` container, except using:
```yaml
    command: [
      "celery",
      "--app", "miqa.celery",
      "beat",
      "--loglevel", "INFO"
    ]
```

That gives developers the ability (e.g. via `docker-compose up django celery`) to run only the Celery worker, without the beat service, which can sometimes be nice for debugging.

miqa/celery.py
```diff
@@ -14,3 +16,11 @@
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+if settings.DEMO_MODE:
+    app.conf.beat_schedule = {
+        'reset-demo': {
+            'task': 'miqa.core.tasks.reset_demo',
+            'schedule': timedelta(days=1),
```
brianhelba: You might want more control of the exact time at which it executes, rather than just specifying the interval between executions. Consider using a [Celery crontab schedule](https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#crontab-schedules) like:
```suggestion
            'schedule': crontab(minute=0, hour=0),  # daily at midnight
```

You'll also need:
```python
from celery.schedules import crontab
```

miqa/core/tasks.py
```diff
@@ -49,6 +49,21 @@ def _download_from_s3(path: str, public: bool) -> bytes:
...
+    demo_project.import_path = 's3://miqa-storage/miqa.csv'
+    demo_project.export_path = 'samples/demo.json'
+    demo_project.save()
+    import_data(demo_project.id)
```
brianhelba, line 58: If it's important that this be executed synchronously, maybe add a code comment indicating why.

Naively, I imagine this could have been written as:
```python
import_data.delay(demo_project.id)
```
but I also suspect there could be important reasons to ensure the demo project data gets re-imported ASAP.

brianhelba (approved):
All this LGTM, and I think it'll work exactly as intended to support daily resets of the demo project.

I've included a number of suggestions inline to possibly make the code a bit more maintainable, but of course use your own judgement as to whether you like each suggestion and its associated trade-offs. I don't know very much about the development context of MIQA.

miqa/core/tasks.py
annehaley: Removed in [418868c](https://github.com/OpenImaging/miqa/pull/635/commits/418868c2527eca6e3bbd4f287477a48908834cce)

annehaley: Removed in [418868c](https://github.com/OpenImaging/miqa/pull/635/commits/418868c2527eca6e3bbd4f287477a48908834cce)

miqa/core/tasks.py
annehaley: Removed in [418868c](https://github.com/OpenImaging/miqa/pull/635/commits/418868c2527eca6e3bbd4f287477a48908834cce)

annehaley: Removed in [418868c](https://github.com/OpenImaging/miqa/pull/635/commits/418868c2527eca6e3bbd4f287477a48908834cce)

miqa/celery.py
annehaley: Changed in [5a36a70](https://github.com/OpenImaging/miqa/pull/635/commits/5a36a70f895942267b886b82c0ed7cabb62d4a76)

miqa/celery.py
annehaley: Changed in [5a36a70](https://github.com/OpenImaging/miqa/pull/635/commits/5a36a70f895942267b886b82c0ed7cabb62d4a76)

PR #636, opened 2022-11-21
Multiple decisions

annehaley:
Kilian recently requested support for multiple decisions being saved to an export file and read during import.
This is the one-hour version of adding this feature. It is only supported for json-format export files.

### DIFF
--- a/miqa/core/conversion/import_export_csvs.py
+++ b/miqa/core/conversion/import_export_csvs.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+import pandas
 from pathlib import Path
 from typing import List, Optional as TypingOptional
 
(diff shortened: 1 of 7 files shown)


miqa/core/conversion/import_export_csvs.py
```diff
@@ -63,6 +65,16 @@ def validate_import_dict(import_dict, project: TypingOptional[Project]):
                                     Optional('session_id'): Or(str, None),
                                     Optional('scan_link'): Or(str, None),
                                     'frames': {And(Use(int)): {'file_location': And(str)}},
+                                    Optional('decisions'): [
+                                        {
+                                            'decision': And(str),
```
zachmullen: Does schema need `And` to wrap a single type?

zachmullen (approved):

PR #637, opened 2022-11-25
Minor/Simple Code Tweaks/Fixes

davidshq:
(ordered roughly from most to least signficant):
1. Vuetify auto-loads its own css files, so loading `vuetify.min.css` and `materialdesignicons.min.css` in `main.ts` causes stylesheets to be loaded twice
2. The hex codes for the `color` property of CSS should be unquoted
3. Vuetify now uses `text-h6` instead of `title`, `text-h5` instead of `headline`
4. Only using units on CSS styles when a non-zero value has been specified is currently considered best practice

### DIFF
--- a/web_client/src/components/DataImportExport.vue
+++ b/web_client/src/components/DataImportExport.vue
@@ -187,7 +187,7 @@ export default defineComponent({
         >
           <v-icon>mdi-close</v-icon>
         </v-btn>
-        <v-card-title class="title">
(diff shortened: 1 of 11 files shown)


annehaley (approved):

curtislisle:
Ready for me to fork the MIQA repo? On Mar 29, 2023, at 10:12 AM, Anne Haley ***@***.***> wrote:﻿
Merged #637 into master.

PR #639, opened 2022-12-08
Move definition of celery beat schedule

annehaley:
The demo instance hasn't been running the reset task. Moving the definition fixed this in a dev env.

### DIFF
--- a/miqa/settings.py
+++ b/miqa/settings.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 from pathlib import Path
 
-from celery.schedules import crontab
 from composed_configuration import (
     ComposedConfiguration,
     ConfigMixin,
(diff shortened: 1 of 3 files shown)


miqa/settings.py
```diff
@@ -46,7 +46,17 @@ class MiqaMixin(ConfigMixin):
...
     ACCOUNT_FORMS = {'signup': 'miqa.core.rest.accounts.AccountSignupForm'}
 
-    CELERY_BEAT_SCHEDULE = {}
+    @property
+    def CELERY_BEAT_SCHEDULE(self):
+        if self.DEMO_MODE:
+            return {
+                'reset-demo': {
+                    'task': 'miqa.core.tasks.reset_demo',
+                    'schedule': timedelta(days=1),
```
zachmullen, line 55: This also changes from `crontab` to `timedelta` -- is that a necessary part of this fix?

annehaley: I couldn't get the task to run with crontab; I only observed it working with timedelta

annehaley:
Also, do you know why the lint test is getting the current error?

zachmullen:
No idea, but I assume it's our use of cache-tox in CI. Maybe try removing or upgrading that.

zachmullen (approved):

PR #640, opened 2022-12-12
Several fixes for new deployment

zachmullen:
Each of these commits represents an independent fix. I tested the ansible changes locally and it seems to be working.


annehaley (approved):
Looks good! I didn't know about the `lineinfile` trick, nice!

PR #641, opened 2022-12-13
Allow LFS checkout in heroku push

annehaley:


zachmullen (approved):

zachmullen:
I have no idea if this will work, but we can try. If not, let's just ditch LFS and inline the files.

PR #642, opened 2023-01-05
Fix updating window on frame change

annehaley:
When switching between images, the automatic range values were computing to the same min and max because the distribution value was not updating to the new image. Making the image data and distribution data _computed values_ ensured that they change when the image changes.

### DIFF
--- a/web_client/src/store/index.ts
+++ b/web_client/src/store/index.ts
@@ -51,7 +51,9 @@ function prepareProxyManager(proxyManager) {
       view.getRepresentations().forEach((representation) => {
         representation.setInterpolationType(InterpolationType.NEAREST);
         representation.onModified(macro.debounce(() => {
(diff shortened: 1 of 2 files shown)


web_client/src/store/index.ts
```diff
@@ -51,7 +51,7 @@ function prepareProxyManager(proxyManager) {
       view.getRepresentations().forEach((representation) => {
         representation.setInterpolationType(InterpolationType.NEAREST);
         representation.onModified(macro.debounce(() => {
-          view.render(true);
+          if (view.getRepresentations()) view.render(true);
```
zachmullen: Let's always use curly braces with `if` statements (if possible we can add this to our eslint config, or whatever tool we are using for that now)

zachmullen (approved):

PR #650, opened 2023-05-01
Fix window lock bug

annehaley:
The cause for the bug was that we failed to save window changes to the state.


annehaley:
@zachmullen Could you investigate why the Cloudflare tests are failing here and on master? I don't have access to these logs.

zachmullen:
Seems like a bug on CloudFlare's side, this was the log

```
15:06:03.091 | Cloning repository...
-- | --
15:06:07.109 | From https://github.com/OpenImaging/miqa
15:06:07.109 | * branch            614118423f05dbca8f5d7756e94192865c16c41b -> FETCH_HEAD
15:06:07.110 |  
15:06:07.667 | HEAD is now at 6141184 Fix window lock bug
15:06:07.668 |  
15:06:07.785 |  
15:06:07.815 | Success: Finished cloning repository files
15:13:19.669 | Failed: an internal error occurred
```

I restarted the build manually.

zachmullen:
The retried build also failed. I think there's a 25MB file size limit on CloudFlare Pages... are any of our bundles larger than that? Thinking in particular of wasm bundles, if we are including those on the deployed site.

zachmullen (approved):

PR #657, opened 2023-06-15
Upgrade vtk.js version

annehaley:


web_client/src/store/index.ts
```diff
@@ -227,9 +229,11 @@ function progressHandler(completed, total) {
 }
 
 /** Creates array of tasks to run then runs tasks in parallel. */
-function startReaderWorkerPool() {
+function startReaderWorkerPool(onDownloadProgress) {
   // Get the current array of tasks in readDataQueue
-  const taskArgsArray = readDataQueue.map((taskInfo) => [taskInfo]);
+  const taskArgsArray = readDataQueue.map((taskInfo) => [
+    Object.assign(taskInfo, { onDownloadProgress }),
```
zachmullen: Should we do the spread operator here, or do we have to do `Object.assign` to modify in-place?

web_client/src/types.ts
```diff
@@ -183,6 +183,7 @@ interface MIQAStore {
   currentFrameId: string | null;
   loadingFrame: boolean;
   errorLoadingFrame: boolean;
+  onDownloadProgress: (e: Event) => null | null;
```
zachmullen: Should this be `(e: Event) => void` instead? (I am not a typescript expert so I'm not sure)

zachmullen (approved):
LGTM

web_client/src/store/index.ts
annehaley: The spread operator suffices here, so I changed it in [ded9674](https://github.com/OpenImaging/miqa/pull/657/commits/ded9674a438a8b0663cb19b748faaa0dcf107ce8)

web_client/src/types.ts
annehaley: Yes, you're right. I changed it in [9bb7713](https://github.com/OpenImaging/miqa/pull/657/commits/9bb7713fe1ad7c76d76d54829f5afb1b73072fdf)

PR #658, opened 2023-12-04
Fix reset demo logic

zachmullen:
Under the old logic, if the Demo Project was deleted or renamed, the reset would fail.

Also, the import file path on S3 was incorrect. This was also fixed.


annehaley (approved):
LGTM, thanks for taking care of this.

